PR #22, opened 2026-08-19
fix: base64 padding emitted before the data, corrupting auth headers (issue #15)

### LINKED ISSUE #15: HTTP fetch path drops the userinfo (user:pass@) component — basic auth is never sent
The HTTP client discards the credential component when normalizing the URL,
so every tier-1 request goes out unauthenticated. Visible in the response:
the echoed `url` comes back without the userinfo it was given.

    web_fetch url=https://user:pass@authenticationtest.com/HTTPAuth/ tier=1
      → http-fetch AuthWall status=401
      → url: "https://authenticationtest.com/HTTPAuth/"   ← credentials gone

Chromium honors userinfo natively, so the browser tier authenticates fine —
which is what makes the failure confusing rather than obvious:

    tier=2  → ghost-render captcha=false dom=3KB, verdict ContentOk
            → http-retry-with-ghost-cookies (tier 1) status=401
            → 401 returned, the rendered page is discarded

That retry inherits the same defect: it carries ghost cookies but no
credentials, so it re-hits a wall the browser had already cleared, and its
401 replaces content that was already in hand.

Consequences:
- tier auto/1 can never fetch a basic-auth URL.
- tier 2 fails too, despite having the content.
- `next_action` says "no extractable content behind the wall — use an
  interactive agent browser", contradicting the `dom=3KB` line above it.

Workaround — adding any `actions` step pins execution to the browser and
skips the tier-1 retry entirely. A no-op `wait` is enough:

    web_fetch url=https://user:pass@authenticationtest.com/HTTPAuth/ \
              tier=2 \
              actions='[{"do":"wait","ms":200}]'

      → escalation: browser-launch ok
                    ghost-render captcha=false dom=3KB
                    actions "1 steps ok"           ← no tier-1 retry
      → status 200, tier "2-actions", content_ok true

      # Login Success
      Success! You are now logged in! ...

The wait does nothing functionally — its only effect is routing. Same URL,
same tier, opposite outcome, which isolates the retry as the failure point.

Fix: carry userinfo through as an `Authorization: Basic` header on the HTTP
path. That makes tier 1 work directly and removes the tier-2 regression as a
side effect.

Repro host is public: https://authenticationtest.com/HTTPAuth/ (user/pass).

Mart-Bogdan:
## Summary

The encoder decided padding from the bit-shift instead of the output position: `pad = chunk.len() * 8 < shift + 6`, where `shift = 18 - 6*i` counts down as `i` counts up. So for a partial group the test was inverted in order — a 1-byte chunk padded at i=0 and i=1, then emitted its data at i=2 and i=3. `user:passwd` came out as `dXNlcjpwYXNz==QA` instead of `dXNlcjpwYXNzd2Q=`.

Output char `i` encodes bits [i*6, i*6+6), so it is padding exactly when the chunk has no bits that far in: `pad = i * 6 >= chunk.len() * 8`.

Credentials that were an exact multiple of 3 bytes long encoded correctly, which is why this survived the #15 fix — every other length went out mangled and was rejected as unauthenticated.

Adds RFC 4648 §10 vectors and a credentials regression test.

Fixes #15

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix #15
- [ ] `test` — tests only

## Checks

- [x] `cargo test --release` passes (496 passed; 0 failed;)
- [ ] `cargo clippy --release -- -Dwarnings` passes (6 pre existing)

## Breaking changes

This logic also affect proxy, but it was same way broken

## Test plan

fetch https://user:passwd@httpbin.io/basic-auth/user/passwd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.3.7] - 2026-08-19
+
+### Fixed
+
+- **Basic auth and proxy auth headers were corrupted by a base64 bug (#15)**: the encoder placed its `=` padding at the start of the final group instead of the end, so `user:passwd` encoded as `dXNlcjpwYXNz==QA` rather than `dXNlcjpwYXNzd2Q=`. Only credentials whose byte length was an exact multiple of 3 came out valid; everything else was rejected by the server. Covered by RFC 4648 test vectors.
+
 ## [2.3.6] - 2026-08-19
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -807,7 +807,7 @@ dependencies = [
 
 [[package]]
 name = "donsetch"
-version = "2.3.6"
+version = "2.3.7"
 dependencies = [
  "boring",
  "boring-sys",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "donsetch"
-version = "2.3.6"
+version = "2.3.7"
 edition = "2024"
 license = "AGPL-3.0"
 description = "Web fetch, search and crawl for AI agents. Custom stealthy HTTP fetch tier on BoringSSL."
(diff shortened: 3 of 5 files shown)


dondai44423:
Merged via cherry-pick (commit 9ef55ed) with conflict resolution. The base64 padding fix is in master. You get contributor credit.

PR #23, opened 2026-08-19
fix: isolate PDF extraction on aarch64 to prevent hang

p-coding:
Fixes linux-aarch64 20m12s timeout on \`cargo test --target aarch64-unknown-linux-gnu\`.

Root cause \`src/crawl/tests.rs:1021\` hand-rolled PDF had bad Length/xref. On aarch64 PDFium holds global \`OnceLock<Mutex<PdfiumCore>>\` \`src/pdf/engine.rs:120\` inside tokio worker and hangs.

Changes:
- \`src/pdf/engine.rs\`: quick_validate + try_core_timeout(3s) before lock
- \`src/crawl/mod.rs\`: PDF bodies to spawn_blocking + timeout 3s
- \`src/crawl/tests.rs\`: minimal_pdf helper

Test: 484 passed (was 483 + 1 hung >60s).

### DIFF
--- a/src/crawl/mod.rs
+++ b/src/crawl/mod.rs
@@ -795,7 +795,42 @@ impl Crawler {
                     } else {
                         ctype
                     };
-                    let mut r = match extract::extract(body_bytes, body_ctype, &page_url, &eo) {
+                    // ── ARM64 PDFium hang isolation ──
+                    // `extract::extract` -> `pdf::parse` -> `engine::load_document`
+                    // holds a global `Mutex<PdfiumCore>` and calls blocking
+                    // `FPDF_RenderPageBitmap`. On aarch64 a malformed PDF can
+                    // hang indefinitely while holding the lock, stalling the
+                    // tokio worker thread. Offload PDF bodies to the blocking
+                    // pool with a 10s timeout; HTML stays on the fast path.
+                    let is_pdf = body_bytes.starts_with(b"%PDF-")
+                        || body_ctype.to_ascii_lowercase().contains("pdf");
+                    let extract_res: Result<crate::extract::Extracted, crate::extract::ExtractError> =
+                        if is_pdf {
+                            if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                                let body_owned = body_bytes.to_vec();
+                                let ctype_owned = body_ctype.to_owned();
+                                let url_owned = page_url.clone();
+                                let eo_owned = eo.clone();
+                                let task = handle.spawn_blocking(move || {
+                                    extract::extract(&body_owned, &ctype_owned, &url_owned, &eo_owned)
+                                });
+                                match tokio::time::timeout(Duration::from_secs(3), task).await {
+                                    Ok(Ok(Ok(r))) => Ok(r),
+                                    Ok(Ok(Err(e))) => Err(e),
+                                    Ok(Err(join_err)) => Err(crate::extract::ExtractError::BadSelector(
+                                        format!("extract task failed: {join_err}"),
+                                    )),
+                                    Err(_) => Err(crate::extract::ExtractError::BadSelector(
+                                        "extract timed out after 3s".into(),
+                                    )),
+                                }
+                            } else {
+                                extract::extract(body_bytes, body_ctype, &page_url, &eo)
+                            }
+                        } else {
+                            extract::extract(body_bytes, body_ctype, &page_url, &eo)
+                        };
+                    let mut r = match extract_res {
                         Ok(r) => r,
                         Err(e) => {
                             skipped
--- a/src/pdf/engine.rs
+++ b/src/pdf/engine.rs
@@ -133,6 +133,254 @@ pub fn core() -> MutexGuard<'static, PdfiumCore> {
     .unwrap_or_else(|e| e.into_inner())
 }
 
+/// Try to acquire the global PDFium lock with a short timeout, returning
+/// `None` if a previous PDFium call is hung (the lock is held forever by a
+/// leaked blocking thread). This prevents a single hung PDF from
+/// dead-locking all future PDF extractions.
+fn try_core_timeout(dur: std::time::Duration) -> Option<MutexGuard<'static, PdfiumCore>> {
+    let core = CORE.get()?;
+    let start = std::time::Instant::now();
+    loop {
+        if let Ok(g) = core.try_lock() {
+            return Some(g);
+        }
+        if start.elapsed() >= dur {
+            return None;
+        }
+        std::thread::sleep(std::time::Duration::from_millis(20));
+    }
+}
+
+/// Lightweight pre-flight check before touching PDFium.
+///
+/// Malformed Length / xref / startxref values can make PDFium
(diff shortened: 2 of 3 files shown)


src/crawl/mod.rs
```diff
@@ -795,7 +795,42 @@ impl Crawler {
...
+                        if is_pdf {
+                            if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                                let body_owned = body_bytes.to_vec();
+                                let ctype_owned = body_ctype.to_owned();
+                                let url_owned = page_url.clone();
+                                let eo_owned = eo.clone();
+                                let task = handle.spawn_blocking(move || {
+                                    extract::extract(&body_owned, &ctype_owned, &url_owned, &eo_owned)
+                                });
+                                match tokio::time::timeout(Duration::from_secs(3), task).await {
```
Mart-Bogdan, line 817: Your comment at line 804 told 10s timout, but it is 3s. This can also fail on big valid PDF files.

src/crawl/mod.rs
```diff
@@ -795,7 +795,42 @@ impl Crawler {
...
                     };
-                    let mut r = match extract::extract(body_bytes, body_ctype, &page_url, &eo) {
+                    // ── ARM64 PDFium hang isolation ──
+                    // `extract::extract` -> `pdf::parse` -> `engine::load_document`
+                    // holds a global `Mutex<PdfiumCore>` and calls blocking
+                    // `FPDF_RenderPageBitmap`. On aarch64 a malformed PDF can
+                    // hang indefinitely while holding the lock, stalling the
+                    // tokio worker thread. Offload PDF bodies to the blocking
+                    // pool with a 10s timeout; HTML stays on the fast path.
+                    let is_pdf = body_bytes.starts_with(b"%PDF-")
```
Mart-Bogdan, line 805: This duplicates the PDF detection in:
https://github.com/dondai44423/donsetch/blob/3e625050cdd296619eb19ae8e65efc64df85a11e/src/extract/mod.rs#L272

Could it be pulled out into a `pub(crate) fn is_pdf_body(body: &[u8], content_type: &str) -> bool` and reused in both places?

The two are codependent by nature — this guard is only correct as long as it matches what the extractor will actually route to PDFium — so sharing one function makes that structural instead of a convention. They've already drifted slightly (`len() >= 5`, `to_lowercase` vs `to_ascii_lowercase`), harmlessly for now.

src/pdf/engine.rs
```diff
@@ -348,7 +596,13 @@ pub fn rasterize_pages(
     if bytes.len() < 5 || !bytes.starts_with(b"%PDF-") {
         return Err(LoadError::NotPdf);
     }
-    let _lock = core();
+    if let Err(e) = quick_validate(bytes) {
+        return Err(e);
+    }
+    let _lock = match try_core_timeout(std::time::Duration::from_secs(3)) {
+        Some(g) => g,
+        None => return Err(LoadError::Corrupt(super::sys::FPDF_ERR_FORMAT)),
```
Mart-Bogdan, line 604: `LoadError::Corrupt(FPDF_ERR_FORMAT)` here says the document is malformed when it isn't — the file is fine, the engine is unavailable. Same at `src/pdf/engine.rs:676` in `load_document`.

Could this get its own variant, e.g. `LoadError::EngineUnavailable`, so the caller can report what actually happened? In a long-lived process (the MCP server) a leaked lock is terminal — every later PDF fails until restart, and each one silently spends 3s polling first. An unconditional `eprintln!` here would match the convention used elsewhere for failures whose consequences are otherwise invisible.

One caveat on wording: this branch can't currently distinguish a hung engine from ordinary contention — `rasterize_pages` legitimately holds the lock for a full OCR pass, so a concurrent worker can hit the same 3s timeout with nothing wrong. If the error should say "hung, restart required", the abandoned-task case needs marking explicitly (e.g. an `AtomicBool` set when a blocking call is dropped mid-flight). Otherwise "engine unavailable" keeps it honest.

src/crawl/mod.rs
```diff
@@ -795,7 +795,42 @@ impl Crawler {
...
+                                    extract::extract(&body_owned, &ctype_owned, &url_owned, &eo_owned)
+                                });
+                                match tokio::time::timeout(Duration::from_secs(3), task).await {
+                                    Ok(Ok(Ok(r))) => Ok(r),
+                                    Ok(Ok(Err(e))) => Err(e),
+                                    Ok(Err(join_err)) => Err(crate::extract::ExtractError::BadSelector(
+                                        format!("extract task failed: {join_err}"),
+                                    )),
+                                    Err(_) => Err(crate::extract::ExtractError::BadSelector(
+                                        "extract timed out after 3s".into(),
```
Mart-Bogdan, line 824: Following on from the comment in `engine.rs`.

This arm knows something `engine.rs` cannot: the blocking task has been **abandoned**. `tokio::time::timeout` does not cancel `spawn_blocking` — the thread keeps running and, if it is inside PDFium, holds the global core lock forever. So this is the right place to set the flag, and to warn unconditionally via `eprintln!`.

One caveat on what we can claim here: reaching this arm only proves the extraction exceeded 3s, not that the engine hung. A scanned PDF is OCR'd at up to 25 pages at 150 DPI, which legitimately takes much longer — so as written this will fire on healthy documents, and a "restart required" message would be a false alarm.

Suggestion: make the budget large enough (or configurable) that it cannot expire on legitimate OCR work, and treat the two signals together — flag set here ("a PDF task was abandoned; the core lock may now be held forever") plus `try_core_timeout` failing afterwards. That combination justifies telling the user the PDF engine is wedged and only a restart recovers it. A lock timeout on its own just means busy.

Mart-Bogdan (commented):
Not a maintainer, just another contributor passing through -- feel free to disregard. 

Serializing PDFium behind a timeout rather than letting it stall the executor makes sense to me.

Three things I noticed while reading:

- The 3s budget may be tight for valid work -- OCR of a scanned PDF rasterizes up to 25 pages at 150 DPI, which can take much longer, so healthy documents might get dropped as timeouts. The inline comment also says 10s while the code uses 3s.
- On timeout the caller gets "corrupt", though the file is fine and the engine is simply unavailable -- possibly worth its own error variant.
- The PDF detection looks like a copy of `extract/mod.rs:272`; a shared helper would stop the two drifting.

Left details inline.

src/crawl/mod.rs
Mart-Bogdan, line 817: One follow-up question: if this is only a genuine concern on linux-aarch64, could the timeout be gated to that target — or given a much larger window elsewhere?

Probably still having timeout but with way larger budget would be better, than having gate that is tested only on single palatiform.

A generous universal budget (a couple minutes, say) costs nothing on healthy documents, still prevents an unbounded stall everywhere, and keeps one code path under test.

dondai44423:
Merged via cherry-pick (commit 3e6f740). Kept the try_core_timeout and spawn_blocking isolation. Removed quick_validate calls (false-positived on valid PDFs) and fixed try_core_timeout to use get_or_init. You get contributor credit.

PR #52, opened 2026-08-26
fix: harden MCP fetch and browser boundaries

amitamit10:
## Summary

Harden DonSeTch's MCP web-research boundary against SSRF, unsafe browser navigation, cookie-domain confusion, writable-path escape, and unverified release artifacts. The changes keep the default MCP surface keyless and stdio-only while making network and filesystem decisions fail closed.

## Type

- [x] `fix` — bug fix
- [x] `test` — tests included
- [ ] `feat` — new feature

## What changed

- Centralized URL validation for fetch, crawl, MCP URL resolution, browser navigation, redirect handling, and CDP request interception.
  - Rejects non-HTTP(S) schemes, URL credentials, loopback/private/link-local/multicast/reserved addresses, IPv4-mapped IPv6, malformed hosts, and DNS failures.
  - Re-checks browser requests before they are continued, including redirects.
- Uses the Public Suffix List for cookie Domain validation and exact/dot-boundary matching.
- Keeps Chrome sandboxing enabled by default and makes the unsafe opt-in explicit.
- Constrains screenshots, browser output, cache, profile, and debug paths to the DonSeTch state tree with traversal/symlink checks.
- Fails closed on unpinned PDFium assets across all supported target pairs.
- Stops accepting an arbitrary pre-existing npm binary and removes trust in same-release checksum sidecars for self-update; reviewed source pins are required.
- Adds regression tests for the security boundaries and self-update pin behavior.

## Checks

- [x] `git diff --check`
- [x] `cargo check --locked --tests`
- [x] `cargo test --lib` — 660 passed
- [x] `cargo test --tests` — integration soak and token invariants passed
- [x] `cargo clippy --all-targets -- -D warnings`
- [ ] CI is green on all 3 platforms

The local ARM64 run uses the repository's default feature set; the upstream matrix remains responsible for the feature-enabled x86_64/macOS/Windows builds.

## Security notes

This is defense in depth for a local stdio MCP server. DNS rebinding cannot be eliminated completely by application-level preflight checks alone; deployments that require protection against a hostile resolver should also enforce egress/network policy.

The self-update pin table intentionally requires a reviewed source change when a new release is published. Missing pins fail closed rather than trusting a checksum file served beside the artifact.

## Breaking changes

- Private/internal URLs, URL credentials, non-HTTP(S) schemes, and DNS failures are rejected instead of being fetched.
- Self-update stops when the release/platform pair has no reviewed source pin.
- Screenshot/debug paths outside the dedicated state root are rejected.

## Test plan

- Exercise URL guard unit tests for loopback, RFC1918, link-local, multicast, reserved, mapped IPv6, credentials, malformed URLs, and DNS failures.
- Exercise cookie tests for public suffix rejection, host-only cookies, parent domains, and dot-boundary matching.
- Exercise screenshot path tests for traversal and symlink escapes.
- Exercise browser sandbox argument tests.
- Exercise the self-update pin table for known and unknown release/platform pairs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -823,6 +823,7 @@ dependencies = [
  "ndarray",
  "oar-ocr",
  "ort",
+ "psl",
  "regex",
  "reqwest",
  "rustls-native-certs",
@@ -2406,6 +2407,21 @@ dependencies = [
  "quote",
  "syn 2.0.119",
 ]
+
+[[package]]
+name = "psl"
+version = "2.1.226"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7bc88482eea924ca3a2f56a547454169af58deef35567965eb4fc2392a834841"
+dependencies = [
+ "psl-types",
+]
+
+[[package]]
+name = "psl-types"
+version = "2.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33cb294fe86a74cbcf50d4445b37da762029549ebeea341421c7c70370f86cac"
 
 [[package]]
 name = "pulp"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ tar = "0.4"
 semver = "1"
 # Agent CLI: spec-driven command generation (builder API, no derive).
 clap = { version = "4", default-features = false, features = ["std", "help", "usage", "error-context", "suggestions"] }
+psl = { version = "=2.1.226" }
 # Semantic reranker — cross-encoder via ONNX Runtime, shares ort with OCR.
 ort = { version = "=2.0.0-rc.12", default-features = false, features = ["std", "ndarray", "download-binaries", "tls-native", "copy-dylibs"], optional = true }
 tokenizers = { version = "0.23", default-features = false, features = ["onig"], optional = true }
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Works with **Claude Code**, **Cursor**, **OpenCode**, **Pi**, **Windsurf**, and 
 npm install -g donsetch
 ```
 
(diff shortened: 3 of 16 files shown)


dondai44423:
Merged with modifications. The security hardening is excellent and has been integrated into master (commit 229601a).

**What was kept (all the security improvements):**
- Centralized SSRF guard with async DNS resolution (guards.rs)
- CDP Fetch.requestPaused interception in the ghost browser (cdp.rs)
- Cookie domain validation with Public Suffix List (cookies.rs)
- Chrome sandbox opt-in via DONGHOST_NO_SANDBOX env var (ghost/mod.rs)
- Screenshot path validation with traversal/symlink checks (paths.rs)
- PDFium fail-closed hash verification (build.rs)
- Post-action navigation re-check (server.rs)
- All 30 new regression tests

**What was reverted and why:**
1. npm/install.js: The empty PINNED_SHA256 map would break all npm installs. The sidecar SHA256 approach is standard practice (npm, pip, cargo all use same-origin checksums). If GitHub releases are compromised, source-controlled hashes are also compromised since the attacker has repo write access. Kept the existing working approach.
2. src/cli/update.rs: The hardcoded pin table for self-update would require manually adding hashes for every future release. Same reasoning as above.
3. src/transport/tcp.rs: The PR removed DONSETCH_ALLOW_PRIVATE_EGRESS, breaking legitimate local-egress use cases (testing, internal docs).
4. README.md / npm/README.md: Removed the npm-disabled notes since npm install still works.

The expanded is_ssrf_ip from guards.rs is still used by the transport layer (the escape hatch now filters with the expanded range).

Thank you for the thorough security work.

amitamit10:
Thanks for the careful review and for integrating the hardening into master. Keeping the SSRF/DNS protection, CDP Fetch interception, Public Suffix List cookie validation, Chrome sandbox defaults, path traversal/symlink checks, PDFium verification, post-action navigation re-check, and the regression tests is a strong outcome. The rationale for reverting the npm/self-update pinning and private-egress changes is clear; separating those into focused follow-up work makes sense. Appreciate the detailed feedback and the merge.

Issue #48, opened 2026-08-26
[bug] tier-2 ghost: Page.navigate CDP timeout on macOS, even on example.com (Chrome for Testing 152)


p4mpus:
## Summary

On macOS, the tier-2 ghost browser launches and connects to CDP, but **every `Page.navigate` times out** — including on `https://example.com`, the most trivial URL. This makes the entire bot-wall / JS-rendering path unusable on macOS.

## Environment
- macOS 26.5.2 (arm64)
- donsetch 3.2.2 (darwin-arm64 binary, SHA256 verified)
- Chrome for Testing 152.0.7977.42

## Reproduction
```
export DONGHOST_CHROME="/path/to/Google Chrome for Testing"
donsetch fetch https://example.com --tier 2
```
Output:
```
[ghost_escalate] first attempt failed: ghost: cdp timeout: Page.navigate, retrying...
[fetch] error: browser automation error: ghost: cdp timeout: Page.navigate
```

Same result on any other URL (e.g. `https://quotes.toscrape.com/js/`).

## What already works
`donsetch doctor` with a real Chrome binary reports:
```
✓ Chrome/Chromium   152.0.0.0 (probed) at .../Google Chrome for Testing
✓ Browser launch    launched in ~600-1100ms, fingerprint clean (webdriver=false)
```
So the ghost binary launches, connects to DevTools, and reports a clean fingerprint — but navigation never returns. Tier-1 (plain HTTP fetch) and the keyless search engine both work fine.

## Suspected cause
The CDP `Page.navigate` call on the attached page target hangs / times out. Possibly a target-attachment or session-routing issue with Chrome for Testing's headless target on macOS, or a missing `Page.enable`/navigation-lifecycle wait. The browser itself is functional (manual CDP attach to the same binary works).

## Notes
- The default auto-detected `/Applications/Chromium.app` on some Macs can be a non-working stub; pointing `DONGHOST_CHROME` at a genuine Chrome/Chromium binary is required first. That part is resolved here.
- `DONGHOST_DEBUG=1` produces no additional diagnostic output on the navigate failure.
- No proxy configured.

## Workaround
None on macOS at present — tier-2 is fully blocked. Tier-1 HTTP fetch + keyless search remain usable.

dondai44423:
on it, will ship a update right away, since i dont have a mac machine, i cant fully verify is live, but it should work, i will try my best to simulate

dondai44423:
[...]

Issue #51, opened 2026-08-26
DonSeTch crashes with Illegal instruction (core dumped) on QEMU-emulated x86_64 (exit 132)


James-Butler2026:
Environment:
- ACEMAGIC M1 (mini PC) → Proxmox → Ubuntu VM → Docker container
- x86_64 (uname -m: x86_64)
- glibc 2.41 (Debian 13)
- donsetch v3.2.3 via npm install -g donsetch (npm 10.9.8)

Steps to reproduce:
1. npm install -g donsetch
2. Run any subcommand: donsetch --help, donsetch mcp, or the direct binary /opt/data/home/.npm-global/lib/node_modules/donsetch/binaries/donsetch fetch https://example.com
3. Process dies immediately with Illegal instruction (core dumped), exit code 132.

Observed:
- Crashes on every command, including --help and mcp mode. No output at all before the crash.
- Binary is a valid ELF64 x86_64 (readelf: Type DYN, Machine AMD X86-64).
- ldd reports no missing libraries.
- Same crash when running the raw binary directly, outside npm.

Suspected cause:
The prebuilt binary appears to use CPU instructions (likely AVX-512 or similar) that are not available/emulated under QEMU/KVM userspace emulation. This setup runs VMs under QEMU; software compiled for newer instruction sets aborts with SIGILL on the first executed instruction. This is a known failure pattern for Rust binaries shipped prebuilt (also affects gbrain reindex in the same environment).

Impact:
The tool is unusable on any QEMU-virtualized/emulated x86_64 setup, which includes common Proxmox/hypervisor setups and some CI runners. It would be helpful if the prebuilt binary targeted a more conservative x86-64 baseline (e.g. x86-64-v2/x86-64-v1), or if a source build option (cargo build --release) was documented as the fallback for such systems.

dondai44423:
Fixed in master (commit 229601a). The release workflow now sets `RUSTFLAGS=-C target-cpu=x86-64` for x86_64 targets, so the prebuilt binary uses only baseline SSE2 instructions instead of AVX2/AVX-512 from the Actions runner. This will take effect in the next release. If you need a working binary right now, building from source with `cargo build --release` on your QEMU host produces a compatible binary since your local CPU (emulated) won't emit unsupported instructions.

PR #54, opened 2026-08-27
fix(ghost): bounded session CDP waits — tier-2 hard failures when first navigation response stalls

Brandon168:
**Target:** dondai44423/donsetch, master (229601a)
**Branch:** fix/debian-cdp-response-deferral → Brandon168 fork
**Title:** fix(ghost): bounded session CDP waits — tier-2 hard failures when first navigation response stalls

---

## Summary

On Debian 12 (container, 2 vCPUs) with Chromium 151.0.7922.173, every
tier-2 fetch failed with `page returned no content` — including
`example.com`. Upstream #48's fix (dispatch-and-poll instead of awaiting
the `Page.navigate` response) is correct as far as it goes, but on this
box the failure shape is broader: **all** session-scoped CDP responses
(`Page.navigate`, `DOM.getDocument`, `DOM.getOuterHTML`) stall until the
browser finishes its first-navigation settle (~26s here), then flush
together; after that, everything answers in milliseconds. Browser-level
commands (`Target.*`) never stall, which is why `doctor` passes while
fetches fail.

The magnitude is almost certainly environment-specific — for calibration,
plain `chromium --headless=new --dump-dom https://example.com` takes
25.6–26.4s wall on this container (measured repeatedly), and a long-running
patchright-driven Chrome on the *same box* handles the same navigation in
0.3–0.7s via CDP. Healthy machines will likely see a much shorter window
or none at all. What is not environment-specific is the **failure mode**:
wherever a first-navigation response stall exceeds the hardcoded ~20s
render window (slow containers, loaded hosts, CI runners), an unbounded
session-scoped DOM read consumes the whole window and turns a recoverable
wait into a hard "no content" failure. The patch makes the outcome
deterministic: the slow case absorbs the cost once per launch instead of
failing every fetch; the fast case pays nothing. #48 plausibly belongs to
the same family — session-command stalls during warmup, different flavor.

## Changes

1. **Launch warmup** (`Ghost::launch`): navigate to a real HTTPS URL once,
   response capped at 35s, tolerated on failure. This absorbs any
   first-navigation deferral at launch time instead of inside the caller's
   20s render window.
   - Must be a *real* URL: about:blank and data: URLs do not trip or absorb
     the deferral (verified).
   - Flows through the same Fetch request guard as any other navigation;
     SSRF posture unchanged.
2. **`navigate()`**: cap the awaited response at 8s and tolerate timeout as
   "still settling" — complements the existing dispatch-and-poll logic;
   failures still surface via the URL poll loop and redirect guard.
3. **`outer_html()`**: bound both legs (getDocument 3s, getOuterHTML 5s)
   so one queued response costs a single poll iteration rather than the
   whole render window.
4. **`Cdp::call_with_timeout`**: explicit per-call timeout plumbing.

## Verification

- 666 tests pass; clippy `-Dwarnings` clean on top of master's strict clippy.
- Live tier-2 on Debian 12 / Chromium 151 — bot.sannysoft.com (10.7KB),
  creepjs (16KB), Cloudflare-fronted Stack Overflow question (15.7KB full
  Q&A) all return ContentOk. Pre-fix, all three consistently returned
  "no content".
- Reproduction method that isolated the cause: bare WebSocket CDP client,
  fire `Page.navigate`, separately time it vs `Target.getTargetInfo` —
  navigate's response lags the URL advance by tens of seconds here while
  browser-level calls answer in ms throughout.
- Cold CLI tier-2 costs ~28–31s total on this box (deferral absorbed once
  per launch); the MCP daemon amortizes launch across calls, so steady
  state is fast.

## Suggested follow-up (not in this PR)

An env knob for the hardcoded ~20s inner poll windows in `ghost_fetch`
(e.g. `DONGHOST_PAGE_TIMEOUT_MS`) would let slow-container deployments tune
without recompiling — happy to send as a follow-up if useful.

Side note: the official release binaries require GLIBC 2.38/2.39 symbols
(`__isoc23_strtol*`, `pidfd_getpid/pidfd_spawnp`) and fail on Debian 12
(glibc 2.36), Ubuntu 22.04-class systems. Building releases on ubuntu-22.04
or targeting musl would widen prebuilt compatibility; verified locally that
a ubuntu-22.04-built binary caps at glibc 2.35 and runs fine.

### DIFF
--- a/.github/workflows/build-debian12.yml
+++ b/.github/workflows/build-debian12.yml
@@ -1,0 +1,72 @@
+# Build a DonSeTch binary compatible with older-glibc distros.
+#
+# Why this exists: official release binaries are produced on newer
+# runners and link against GLIBC 2.38/2.39, so they fail on
+# distributions like Debian 12 (GLIBC 2.36) with:
+#   version `GLIBC_2.38' not found
+# Building on ubuntu-22.04 (glibc 2.35) produces a binary that runs on
+# Debian 12 and anything newer, without musl's other tradeoffs.
+#
+# Also pins LIBCLANG_PATH to the distro libclang: Homebrew-style or
+# very-new LLVM builds can require a newer glibc than the runner base,
+# which makes bindgen fail during boring-sys's build script.
+
+name: build-debian12
+
+on:
+  push:
+    branches: [master]
+    paths: ['.github/workflows/build-debian12.yml', 'src/**', 'Cargo.toml', 'Cargo.lock', 'build.rs']
+  workflow_dispatch:
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: donsetch-linux-x64-debian12
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain (pinned)
+        run: rustup toolchain install 1.98 --profile minimal
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y nasm cmake golang-go libclang-dev clang
+          echo "LIBCLANG_PATH=/usr/lib/llvm-14/lib" >> "$GITHUB_ENV"
+
+      - name: Build release binary
+        run: cargo build --release --locked
+
+      - name: Smoke test (version + doctor network check)
+        run: |
+          ./target/release/donsetch version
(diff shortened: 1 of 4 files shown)


dondai44423:
Merged in 3bdb28a. The bounded CDP wait logic and warmup navigation are excellent. Skipped the build-debian12.yml workflow file since the existing release workflow already handles the x86_64 baseline ISA target (issue #51 fix, RUSTFLAGS=-C target-cpu=x86-64). Thank you for the thorough investigation.

PR #58, opened 2026-08-27
feat(mcp): streamable-HTTP transport with sessions, cancellation, auth

imonlinux:
**Summary**

Adds a second MCP transport: streamable-HTTP. donsetch mcp --http (or DONSETCH_TRANSPORT=http) serves the full tool surface over HTTP — JSON-RPC POST at /mcp, the GET SSE stream and DELETE-session endpoints that strict streamable-HTTP clients (OpenCode among them) require, and GET /health for probes. Both transports dispatch through the same request handler, so tools behave identically to stdio.

- **Sessions:** initialize returns an Mcp-Session-Id header. Clients that echo it get a dedicated cancellation registry (notifications/cancelled aborts in-flight calls, same semantics as stdio); session-less clients share a default registry. Ids are 16-char random base62 tokens, idle sessions are GC'd after 30 minutes, the table is bounded at 1024 with oldest-first eviction, and DELETE /mcp ends a session immediately.
- **Hardening:** optional bearer auth (DONSETCH_HTTP_TOKEN, constant-time compare), per-request timeout (DONSETCH_HTTP_TIMEOUT_SECS, default 300, returns a JSON-RPC error on expiry), and CORS off by default with DONSETCH_HTTP_CORS=1 to opt in — MCP clients are processes, not browsers, and a permissive default would let any webpage in a local browser read responses from a localhost instance.
- **Graceful shutdown:** SIGTERM/SIGINT stops accepting requests, drains in-flight ones, and shuts the daemon down so no Chrome processes are orphaned.
- Deps: axum 0.7 / tower-http 0.5 (0.8 is current; happy to bump if preferred).

**Type**

- [x] feat — new feature

**Checks**

- [x] cargo test --features ocr,rerank passes (686 tests)
- [x] cargo clippy --all-targets --features ocr,rerank -- -Dwarnings passes (zero warnings)
- [x] cargo fmt --all -- --check passes

**Breaking changes**

None. stdio remains the default; HTTP is opt-in via --http / DONSETCH_TRANSPORT=http.

**Test plan**

- 8 unit tests in src/mcp/http.rs: auth gate (token set/unset), constant-time compare, session-id format + uniqueness, registry roundtrip + removal, idle GC.
- End-to-end against a release build over HTTP, driven as a real client: initialize → session echo → tools/list → web_fetch → web_search → S-handle round-trip → fabricated-handle rejection → GET SSE stream → DELETE session → dead-session rejected (12/12).
- Auth gate e2e: no token configured = open; token configured = 401 without or with a wrong bearer, 200 with the right one (4/4).
- Notification-only POSTs return 202 Accepted per the streamable-HTTP spec (verified: not 204).
- Behavioral note from stdio testing (pre-existing, not introduced here): a client that closes stdin immediately after queueing requests drops in-flight responses; real MCP clients hold stdin open.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,9 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio-tungstenite = "0.30"
 futures-util = "0.3"
+# HTTP server for MCP HTTP transport (axum, SSE)
+axum = { version = "0.7", default-features = false, features = ["http1", "tokio", "json"] }
+tower-http = { version = "0.5", features = ["cors", "trace"] }
 libc = "0.2"
 dirs = "6"
 unicode-bidi = "0.3"
--- a/README.md
+++ b/README.md
@@ -231,6 +231,75 @@ Or use `npx` without global install:
 
 Works with Claude Code, Cursor, OpenCode, Pi, Windsurf, and any client that speaks MCP. Three tools: `web_fetch`, `web_search`, `web_crawl`.
 
+### HTTP transport (remote clients and debugging)
+
+DonSeTch also speaks MCP over HTTP (the streamable-HTTP transport:
+JSON-RPC via POST, plus the GET SSE stream and DELETE session end that
+strict clients expect). Both transports dispatch through the same
+handler, so tools behave identically. HTTP is opt-in:
+
+```bash
+# Flags
+donsetch mcp --http --host 0.0.0.0 --port 8765
+
+# Env vars (identical effect; flags win over env when both are set)
+DONSETCH_TRANSPORT=http DONSETCH_HTTP_HOST=0.0.0.0 donsetch mcp
+```
+
+MCP clients connect to `http://localhost:8765/mcp`:
+
+```json
+{
+  "mcpServers": {
+    "donsetch": { "url": "http://localhost:8765/mcp" }
+  }
+}
+```
+
+**Testing with curl:**
+
+```bash
+# Health check (always unauthenticated, for probes)
+curl http://localhost:8765/health
+
+# Test the MCP endpoint
+curl -X POST http://localhost:8765/mcp \
+  -H "Content-Type: application/json" \
+  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
+```
+
+**Sessions and cancellation.** The `initialize` response carries an
+`Mcp-Session-Id` header. Echo it back on subsequent requests to get a
+dedicated cancellation registry: posting `notifications/cancelled`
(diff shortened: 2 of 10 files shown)


Cargo.toml
Mart-Bogdan, line 59: Please register this feature as cargo feature, so we don't pull in all http server dependencies, when we don't use it.

README.md
```diff
@@ -231,6 +231,75 @@ Or use `npx` without global install:
...
+semantics as stdio). Session-less clients share one default registry —
+cancellation still works, but request ids share a namespace. Unknown
+or expired session ids get a 404. Sessions idle for 30 minutes are
+dropped; `DELETE /mcp` with the session header ends one immediately.
+
+```bash
+# Cancel request 42 during a long call (with a session header)
+curl -X POST http://localhost:8765/mcp \
+  -H "Content-Type: application/json" \
+  -H "Mcp-Session-Id: <from initialize>" \
```
Mart-Bogdan, line 284: As far as I know this feature is deprecated from protocol, latest version 2026-07-28 don't define this header,

> Protocol versions 2025-03-26 through [2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports) also used the Streamable HTTP transport, but in a different shape: servers could assign a session via the Mcp-Session-Id header (terminated with HTTP DELETE), clients could open a standalone SSE stream with HTTP GET to receive server-initiated messages, servers could send JSON-RPC requests on SSE streams, and streams were resumable via Last-Event-ID. None of these mechanisms are part of this revision.

I don't dig further into a spec yet, but I think it is bad idea to implement older version which is already deprecated, unless we really-really need it.

Please take a look at https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#earlier-streamable-http-revisions

Mart-Bogdan (changes requested):
Hello. This is interesting feature. Thank you for PR.

In my opinion it requires careful look, into how it is implementd. I didn't have time to take detailed look, but couple things caught my mind:

1. it implements already deprecated version of sepcification.
2. It looks like optional feature that should not be enabled by default for all builds, similar how  ocr and rerank is made.


@dondai44423 how do you think should we take this in? I wouldn't rush such big change and adopt it as "beta" -- don't commit to stability of this feature until we sure it works good and satable.

I mean we can merge, but state in readme that it is experimental feature without much guarantee.


I'll take a look at this PR again when I would get a chance.

dondai44423:
Manually applied in commit 79b807d. The PR had conflicts with the AVX dynamic loading changes in Cargo.toml, so the HTTP transport code was integrated manually. All 8 tests pass, 0 clippy warnings. Great work on the implementation.

PR #59, opened 2026-08-27
feat(docker): production image + compose service (stdio)

imonlinux:
**Summary**

Adds a production Docker image and a compose service running the MCP server on stdio, so MCP clients can use the server without a local Rust toolchain.
- Multi-stage rust:slim builder → debian:trixie-slim runtime on the same glibc generation (the ort-sys prebuilt archives reference glibc 2.38+ symbols); arch-aware Go install for BoringSSL (amd64/arm64 via TARGETARCH); builds all features --locked to Cargo.lock; non-root runtime user (uid 1000).
- **PDFium pre-fetch:** build.rs sha256-pins and downloads the PDFium static archive, but its curl path gives up on a single mid-transfer reset (curl 35 is not retried without --retry-all-errors). The Dockerfile pre-fetches it with aggressive retries (8 attempts, all errors, stall detection) and verifies the same KNOWN_HASHES pins, keeping builds fail-closed on flaky networks.
- **ONNX shared library shipping:** since 3.2.5, ort loads ONNX Runtime dynamically — build.rs builds vendor/onnx/libonnxruntime.so and the binary dlopens it at runtime, searching next to the executable first. The image stages and ships the .so next to the binary so OCR/rerank work in-container (verified: donsetch doctor reports ONNX Runtime — AVX detected, shared library present). On arm64, where no ONNX prebuilt exists, the staged dir is empty, the copy is a no-op, and the features self-disable at runtime instead of failing the build. Only the .so is staged, not the large static archive that also lives in vendor/onnx.
- **Cache volume ownership:** the cache dir is created and chowned in the image *before* the VOLUME declaration — without that, fresh volumes get a root-owned mountpoint and the non-root user can neither write the fetch cache nor download the rerank model on first use.
- DONGHOST_NO_SANDBOX=1 is set by default: Chromium's sandbox cannot initialize inside a default Docker container (no unprivileged user namespaces) and the browser dies before the DevTools handshake. The container boundary + non-root user provide the isolation. This is upstream's documented escape hatch ("only use in isolated containers").
- Optional INSTALL_CHROME=true build arg for tier 2 browser escalation (chromium + xvfb, ~350 MB).
- compose: persistent cache volume, 2 GB memory ceiling, init for zombie reaping, 45 s stop grace period.

Heads-up: this PR and the HTTP transport PR both add an entry under ## [Unreleased] in CHANGELOG.md; whichever merges second gets a trivial 3-line conflict there.

**Type**

- [x] feat — new feature

**Checks**

- [x] cargo test --features ocr,rerank passes
- [x] cargo clippy --all-targets --features ocr,rerank -- -Dwarnings passes
- [x] cargo fmt --all -- --check passes

No Rust sources are touched (Dockerfile, compose, .dockerignore, README, CHANGELOG), but the gates were run on this tree.

**Breaking changes**

None.

**Test plan**

Built on linux/amd64 and validated in-container, driven as a real stdio MCP client (stdin held open, --supervised):

- initialize / tools/list OK (web_fetch, web_search, web_crawl).
- web_fetch https://example.com → 200, tier 1, content_ok: true.
- web_search → results with minted S-handles; web_fetch via S-handle resolves prewarmed in ~6 ms; fabricated handle SZZZZZZ99 rejected.
- donsetch doctor in the image: ✓ ONNX Runtime — AVX detected, shared library present; stderr shows [onnx] Runtime loaded from /usr/local/bin/libonnxruntime.so.
- Fresh cache volume: rerank model + tokenizer download and cache under the non-root user (model cached. tokenizer cached.).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,54 @@
+# Docker ignore patterns for DonSeTch
+
+# Git
+.git
+.gitignore
+.gitattributes
+
+# CI/CD
+.github
+
+# Rust
+target/
+**/target/
+
+# Fuzzing (development-only)
+fuzz/
+
+# Benchmarks (development-only)
+bench/
+
+# Scripts (development-only)
+scripts/
+
+# Assets (not needed for runtime)
+assets/
+
+# npm package (installed separately)
+npm/
+
+# Documentation (not needed for runtime)
+docs/
+*.md
+!README.md
+
+# IDE
+.vscode
+.idea
+*.swp
+*.swo
+*~
+
+# OS
+.DS_Store
+Thumbs.db
+
+# Docker files (don't copy into image)
+docker-compose*.yml
+.dockerignore
+Dockerfile*
+
+# Development files
+*.log
+.env.*
+!.env.example
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to DonSeTch are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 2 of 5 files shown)


Mart-Bogdan:
> , so MCP clients can use the server without a local Rust toolchain
That's not exactly true. You don't need to have Rust toolchain, to use prebuilt binaries of donsetch, but generally Docker image is good idea.

Need to take a look.

PR #60, opened 2026-08-27
feat(search): make reranker threads container-aware

Bnbig:
## Summary

I ran into substantial CPU throttling while testing concurrent searches in a CPU-limited container. The search results were fine, but ONNX Runtime's default intra-op pool was sized from the host rather than the process's effective CPU budget, so the reranker created more runnable work than the container could execute.

This PR makes the reranker container-aware at session initialization:

- read the process's effective parallelism (`available_parallelism`, including Linux cgroup quota and CPU affinity);
- clamp ONNX intra-op threads only when that value is below the host's physical core count;
- preserve ONNX's native default and affinity behavior on unconstrained hosts;
- provide `DONSEEK_RERANK_THREADS` as an explicit cross-platform override;
- reject invalid overrides and fall back to automatic selection;
- recover the session builder and keep the reranker enabled if ONNX rejects the thread setting.

The model, blend weight, candidate set, scoring and result ordering logic are unchanged. `num_cpus` is optional and is only included by the existing `rerank` feature.

## Type

- [x] `feat` — new feature
- [ ] `fix` — bug fix
- [ ] `test` — tests only

## Measured impact

I built the baseline and patched binaries from the same upstream commit with Rust 1.98, `--features ocr,rerank`, the same dependency cache and `-C target-cpu=x86-64`. Both ran in isolated Linux containers on an 8-core host with a 2-CPU quota. The model/runtime were warmed before measurement and the two variants ran simultaneously with separate search caches.

Each round below is eight concurrent `web_search` calls:

| Round | Variant | Wall | CPU time | Cgroup throttled time | Completed |
|---|---|---:|---:|---:|---:|
| 1 | upstream | 9.506 s | 5.589 s | 4.114 s | 8/8 |
| 1 | patched | 6.712 s | 4.079 s | 0.023 s | 8/8 |
| 2 | upstream | 6.644 s | 5.629 s | 3.161 s | 8/8 |
| 2 | patched | 6.346 s | 4.537 s | 0.112 s | 8/8 |

Across the two rounds, the patched build used about 23% less CPU, reduced cgroup throttled time by about 98%, and reduced mean wall time by about 19%. All 32 measured calls completed without errors and returned the same result count.

Public-engine latency and availability still introduce noise: round 1 produced 33 versus 36 engine reports, while round 2 produced 22 for both variants. I therefore consider the CPU and throttling reductions the strongest signal here, rather than presenting this as a search-quality benchmark.

## Container matrix

I also exercised the compiled binary under different Docker cgroup configurations:

| Constraint | Selected behavior |
|---|---|
| `--cpus=0.5` | 1 intra-op thread |
| `--cpus=1` | 1 intra-op thread |
| `--cpus=2` | 2 intra-op threads |
| `--cpus=4` | 4 intra-op threads |
| `--cpuset-cpus=0-1` | 2 intra-op threads |
| no quota / full host affinity | preserve ONNX default |
| 2 CPUs + `DONSEEK_RERANK_THREADS=1` | explicit 1-thread override |
| 2 CPUs + invalid override | warning, then automatic 2-thread selection |

Every matrix case completed initialization, reranking and a real search successfully.

## Checks

- [x] `cargo test --features ocr,rerank` passes

Additional validation:

- Linux release-mode suite: 677 unit tests, 1 soak test and 4 token-invariant tests passed;
- `cargo deny --all-features check --hide-inclusion-graph` passed (advisories, licenses, bans and sources);
- core dependency graph without `rerank` does not include `num_cpus`;
- the runtime matrix above covered automatic selection, explicit override, invalid override fallback, cgroup quota, CPU affinity and the unconstrained path.

## Breaking changes

None. Existing unconstrained installations keep ONNX's default thread and affinity behavior. CPU limits and the environment override are sampled when the reranker session is initialized, so a process should be restarted after an in-place CPU-limit change.

## Test plan

1. Run the full test, Clippy, formatting and supply-chain gates.
2. Start a rerank-enabled build under 0.5, 1, 2 and 4 CPU quotas and verify the initialization log matches the effective budget.
3. Start it with a two-CPU cpuset and verify affinity is respected without a quota.
4. Verify an explicit positive override wins over automatic detection.
5. Verify invalid and non-Unicode values do not disable reranking and fall back safely.
6. Compare same-commit upstream/patched binaries under concurrent searches and record wall time, CPU usage and cgroup throttling.

— Alfonso Ada


dondai44423:
Manually applied in commit 9d32cfd. The PR had conflicts with the ONNX Cargo.toml changes. Thread selection logic integrated with 6 new tests. Great work.

Issue #67, opened 2026-08-28
mcp --http silently starts a stdio server when the binary is built without the http feature


imonlinux:
> The --http branch in src/main.rs is #[cfg(feature = "http")]. In a build without the feature, donsetch mcp --http — and DONSETCH_TRANSPORT=http — fall through to stdio with no error and no warning. One repro against a Docker image built from the current Dockerfile (--features ocr,rerank):
>
> 
```
$ docker run -i --rm donsetch-mcp donsetch mcp --http    # + initialize request on stdin
{"id":1,"jsonrpc":"2.0","result":{"protocolVersion":"2026-07-28","serverInfo":{"name":"donsetch","version":"3.2.5"}}}
```

>
> --http was passed; the server answered initialize **over stdio**. Same happens with DONSETCH_TRANSPORT=http in compose, and with the linux-arm64 / darwin-x64 release binaries. From the outside you can't tell "wrong flag" from "wrong build".
>
> Proposed fix in a small PR: when --http or DONSETCH_TRANSPORT=http is requested and the feature isn't compiled in, exit with an error naming the cargo feature.
>
> Secondary, related: mcp ignores unknown flags generally (--bind, typos, anything). That's why the README's --token fails silently instead of loudly. Would you prefer a hard error on unrecognized mcp args, or a stderr warning? I'd lean hard error — mcp's flag surface is small — but it's a behavior change so I'm asking first.

PR #69, opened 2026-08-28
fix(docs): document the HTTP transport's actual interface

### LINKED ISSUE #66: README: HTTP transport section documents flags and env vars that do not exist
> Hey @dondai44423 — while testing the new HTTP transport I found that the README section documents an interface the binary doesn't have:
>
> | README says | Code has |
> |---|---|
> | donsetch mcp --http --bind 0.0.0.0:8765 | --host HOST / --port PORT (src/main.rs, get_arg_value) |
> | --token "secret" flag | token only via DONSETCH_HTTP_TOKEN env |
> | DONSETCH_HTTP_BIND | DONSETCH_HTTP_HOST / DONSETCH_HTTP_PORT |
> | DONSETCH_HTTP_CORS=* | accepts only 1/true/on (src/mcp/http.rs); when on, it allows any origin |
>
> The practical problem is the --token line: unknown flags are silently ignored, so donsetch mcp --http --bind 0.0.0.0:8765 --token "secret" starts the server on the default 127.0.0.1:8765 **with no auth** while the user believes it's protected. DONSETCH_HTTP_CORS=* also silently does nothing.
>
> The merged code matches what PR #58 proposed (--host/--port, CORS=1); the --bind/--token prose only exists in the README section from the manual integration. route_help() already prints the correct interface for donsetch help mcp, so correct text exists in-tree.
>
> Two smaller stale spots I'd fold into the same fix: the feature-flags build note ("ships with both features" — there are three features now and the prebuilt set is platform-dependent: linux-arm64 and darwin-x64 build core-only), and the ci.yml comment "Release builds ship with http enabled on all platforms" (release.yml shows the two core-only targets).
>
> Fix incoming in a fix(docs) PR. Happy to also add an "experimental" note to the section if you want to take Mart-Bogdan's suggestion from #58 — your call.

imonlinux:
## Summary

Fixes #66.

Fixes the README's HTTP transport section, which documented an interface that doesn't exist: `--bind`, `--token`, `DONSETCH_HTTP_BIND`, and `DONSETCH_HTTP_CORS=*` are not in the code. The real interface is `--host`/`--port` flags and `DONSETCH_HTTP_HOST`/`_PORT`/`_TOKEN`/`_TIMEOUT_SECS` env vars, with `DONSETCH_HTTP_CORS` accepting `1`/`true`/`on` (allow-any when enabled).

The security angle: a user running the documented `donsetch mcp --http --bind 0.0.0.0:8765 --token "secret"` gets a server on the default 127.0.0.1:8765 with no auth — `--bind` and `--token` are silently swallowed. (The unknown-flags behavior that let this drift go unnoticed is the secondary thread in #67 — left open for a maintainer call rather than snuck into a docs PR. The silent stdio fallback itself is #67, with a fix incoming.)

- HTTP section rewritten to the working interface: flags/env, client config, a curl smoke test, sessions & cancellation semantics, the env var table, graceful shutdown.
- New build-requirement note: `http` is an optional cargo feature — source builds need `--features http`, and the linux-arm64/macOS-x64 prebuilts are core-only.
- Feature-flags bullet updated to the three-feature world and which prebuilts carry `http`.
- Gotchas: corrected the "ships with both" row; added a row for `--http` silently falling back to stdio in a build without the feature.
- Docker architecture notes: the image builds `--features ocr,rerank`, not "all features enabled" (marked interim — enabling `http` in the image is a separate PR I'll send).
- ci.yml: corrected the comment claiming release builds ship http "on all platforms" (they do on linux-x64/macOS-arm64/Windows-x64; linux-arm64/macOS-x64 are core-only — hence the corrected notes everywhere else).

Docs only — no code changes.

## Type

- [x] `docs` — documentation fix

## Checks

- [x] `cargo test --features ocr,rerank` passes (no Rust sources touched)

## Breaking changes

None.

## Test plan

- Every flag/env var/endpoint claim verified against `src/main.rs` and `src/mcp/http.rs` at master: `--host`/`--port` resolution order, the `DONSETCH_HTTP_*` env vars, CORS `1`/`true`/`on` matching, session 404s, 30-min idle GC, SIGTERM drain.
- Release-matrix claims read from `.github/workflows/release.yml`: 5 targets, linux-arm64 and macOS-x64 build core-only.
- Both curl examples work verbatim against a running server.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,9 @@ jobs:
             target: x86_64-unknown-linux-gnu
             # http feature tested on Linux only: axum adds 20+ min
             # compile time on Windows MSVC. macOS/Windows CI skip it.
-            # Release builds ship with http enabled on all platforms.
+            # Release builds ship with http on linux-x64, macOS-arm64,
+            # and Windows-x64; linux-arm64/macOS-x64 are core-only
+            # (no ONNX prebuilt for those targets).
             features: --features ocr,rerank,http
             test-skip: ''
             test-mode: ''
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to DonSeTch are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Docs: HTTP transport interface.** The README documented flags and env vars that do not exist (`--bind`, `--token`, `DONSETCH_HTTP_BIND`, `DONSETCH_HTTP_CORS=*`). Replaced with the real interface (`--host`/`--port`, `DONSETCH_HTTP_HOST`/`_PORT`/`_TOKEN`/`_TIMEOUT_SECS`, `DONSETCH_HTTP_CORS=1`), a build-requirement note (`http` is an optional cargo feature; the linux-arm64/macOS-x64 prebuilts are core-only), corrected feature-flag notes, and a Gotchas row for `--http` silently falling back to stdio in a build without the feature.
+
 ## [3.4.0] - 2026-08-28
 
 ### Added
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -43,7 +43,9 @@ jobs:
             target: x86_64-unknown-linux-gnu
             # http feature tested on Linux only: axum adds 20+ min
             # compile time on Windows MSVC. macOS/Windows CI skip it.
-            # Release builds ship with http enabled on all platforms.
+            # Release builds ship with http on linux-x64, macOS-arm64,
+            # and Windows-x64; linux-arm64/macOS-x64 are core-only
+            # (no ONNX prebuilt for those targets).
```
Mart-Bogdan, line 48: I think this slipped i by mistake. Text above tells about http, and ONNX  has nothing to do with it.

imonlinux: Yep, not sure where that came in.

PR #70, opened 2026-08-28
feat(docker): enable the http transport in the image

imonlinux:
## Summary

Refs #67 — the Docker repro there (image built `--features ocr,rerank`, `mcp --http` answering over stdio) goes away at the root once the image builds with the feature.

The image was the only distribution building without the `http` feature — the natural deployment for a remote transport (a container) was the one build that couldn't serve it. This builds the image with `--features ocr,rerank,http --locked`, matching the linux-x64/macOS-arm64/Windows-x64 release binaries.

- `EXPOSE 8765` + a `docker run` HTTP-mode comment; stdio stays the default CMD.
- compose: an opt-in `http` profile (`donsetch-http`) running the same image with `DONSETCH_TRANSPORT=http` and `DONSETCH_HTTP_HOST=0.0.0.0` in-container. The port is published on `127.0.0.1:8765` by default so an unset `DONSETCH_HTTP_TOKEN` never exposes unauthenticated MCP to the LAN — switch to `8765:8765` together with a token to serve remote clients. Listener-based healthcheck (bash `/dev/tcp` — the slim runtime image has no curl/wget). The existing stdio service is untouched.
- README: the Docker section (Option 5) documents the HTTP mode and the profile; architecture notes updated to the three-feature build.

Measured build-time delta from adding axum + tower-http: **3m 43s** on a warm cache (baseline cold release build `--features ocr,rerank`: 6m 53s; no-op rebuild after: 0.45s, so the cache does the work once).

Self-contained — deliberately not stacked on the docs PR for #66 (they touch different README regions, so either can land first). The only shared file region is the CHANGELOG's `[Unreleased]` heading; if #66's PR lands first this may need a one-line de-dup on rebase, happy to do that on request.

## Type

- [x] `feat` — new feature

## Checks

- [x] `cargo test --features ocr,rerank` passes (no Rust sources touched: Dockerfile, compose, README, CHANGELOG)

## Breaking changes

None — stdio remains the default CMD and the default compose flow is unchanged.

## Test plan

Built the image on linux/amd64 (`cargo build --release --features ocr,rerank,http --locked` confirmed in the build log) and drove the compose `http` profile with docker compose v2:

- `docker compose config --profiles` lists `http`; `--profile http up -d donsetch-http` starts **only** the HTTP service — the stdio service is not pulled into `up`.
- `GET /health` → 200 from outside the container; the compose healthcheck flips the container to `healthy`.
- `POST /mcp initialize` → 200 + `Mcp-Session-Id` header; `notifications/initialized` → 202; `tools/list` with the session header → full tool surface.
- Auth gate with `DONSETCH_HTTP_TOKEN` set: no bearer → 401, wrong bearer → 401, right bearer → 200, `/health` stays open.
- `docker compose --profile http down` tears down cleanly; the stdio flow (`docker compose run --rm donsetch`) is untouched by the profile.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,13 +1,19 @@
 # DonSeTch MCP server — docker compose
 #
-# stdio mode, for MCP clients that launch the server as a subprocess
-# (Claude Code, Cursor, etc.):
+# stdio mode (default), for MCP clients that launch the server as a
+# subprocess (Claude Code, Cursor, etc.):
 #
 #   docker compose build
 #   docker compose run --rm donsetch
 #
 # and point your MCP client at the wrapper command:
 #   docker run -i --rm donsetch-mcp donsetch mcp --supervised
+#
+# HTTP mode (the `http` profile below), for remote/debug use:
+#
+#   docker compose --profile http up -d donsetch-http
+#   curl http://localhost:8765/health
+#   MCP clients connect to: http://localhost:8765/mcp
 
 services:
   donsetch:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -56,5 +62,46 @@ services:
         limits:
           memory: 2g
 
+  # ── HTTP transport (opt-in profile) ─────────────────────────────
+  # docker compose --profile http up -d donsetch-http
+  # Serves MCP at http://localhost:8765/mcp (streamable-HTTP).
+  # Same image and cache volume as the stdio service; only the
+  # transport differs. See README → "HTTP transport" for the env vars.
+  donsetch-http:
```
Mart-Bogdan, line 70: **The HTTP service has no `restart:` policy, and it's the one deployment with no crash recovery.** The HTTP branch in [main.rs:29-44](src/main.rs:29) returns before the `--supervised` check, so the crash-only supervisor is stdio-only (the help text says as much). Release builds set `panic = "abort"` ([Cargo.toml:91](Cargo.toml:91)) — the supervisor comment itself says one panicking request would otherwise kill the whole process. In stdio mode the supervisor restarts the child; in this long-lived container, one bad request aborts the process and the container stays down until someone notices.

Probably best way would be to set  `restart: unless-stopped` and don't use `--supervised`, let docker do it things.

docker-compose.yml
```diff
@@ -56,5 +62,46 @@ services:
...
+    profiles: ["http"]
+    image: donsetch-mcp:latest
```
Mart-Bogdan, line 80: This service lacks

```
    build:
      context: .
```
So it would fail to run if image isn't already built locally.
```
PS > docker compose --profile http up -d donsetch-http
[+] up 1/1
 ✘ Image donsetch-mcp:latest Error pull access denied for donsetch-mcp, repository does not exist or may require 'docker login'                                                                                                 1.6s
Error response from daemon: pull access denied for donsetch-mcp, repository does not exist or may require 'docker login'

What's next:
    Debug this Compose error with Gordon → docker ai "help me fix this compose error"
```

Mart-Bogdan (commented):
Your docker compose file won't build image if run with HTTP profole, so docker image must be already built.

Also small problem with supervised

PR #71, opened 2026-08-28
fix(mcp): fail loudly when --http is requested but the feature isn't compiled in

### LINKED ISSUE #67: mcp --http silently starts a stdio server when the binary is built without the http feature

imonlinux:
## Summary

Fixes #67 (the primary ask; the secondary question there — hard error vs warning on unrecognized `mcp` flags generally — is deliberately out of scope pending your call).

In a build without the `http` cargo feature (the linux-arm64 and macOS-x64 prebuilts, and any plain `cargo build`), the transport-selection block in `src/main.rs` compiles out entirely — `donsetch mcp --http` and `DONSETCH_TRANSPORT=http` silently fall through to stdio. A client configured for HTTP then hangs waiting on a listener that never comes up, and nothing on the server side says why.

The `#[cfg(not(feature = "http"))]` counterpart now checks both entry paths and exits with:

```
mcp: HTTP transport requested (--http / DONSETCH_TRANSPORT=http), but this
binary was built without the `http` cargo feature. Rebuild with
--features http, or use a prebuilt binary that includes it.
```

Builds with the feature are unchanged; plain `mcp` and `--supervised` are untouched.

## Type

- [x] `fix` — bug fix

## Checks

- [x] `cargo test --features ocr,rerank` passes

Also gated beyond CI's configs: `cargo clippy --all-targets -- -Dwarnings` on the default (no features) build — the configuration where the guard actually compiles — and with `--features ocr,rerank,http`; `cargo test` on the default build (676 passed).

## Breaking changes

None for binaries that include the feature. A no-feature binary launched with `--http` now exits 1 instead of silently serving stdio — that is the fix.

## Test plan

Behavioral, against dev-profile binaries on linux-x64:

- Default build (no features) + `mcp --http --host 127.0.0.1 --port 9999` → the error above, exit 1.
- Default build + `DONSETCH_TRANSPORT=http mcp` → same, exit 1.
- Default build + plain `mcp` → stdio behavior unchanged (clean EOF exit).
- `--features http` build + `--http` → listener up, `GET /health` → 200; SIGTERM drains in-flight requests and shuts down cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to DonSeTch are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Loud failure for `--http` without the feature.** In a binary built without the `http` cargo feature (the linux-arm64 and macOS-x64 prebuilts, and any plain `cargo build`), `donsetch mcp --http` and `DONSETCH_TRANSPORT=http` silently fell through to stdio — a client configured for HTTP would hang waiting on a listener that never came up. Both paths now exit immediately with an error naming the missing cargo feature and how to get a binary that includes it.
+
 ## [3.4.0] - 2026-08-28
 
 ### Added
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,6 +42,21 @@ async fn main() {
                 }
                 return;
             }
+            // Without the `http` cargo feature the checks above compile
+            // out, so a requested HTTP transport would silently fall
+            // through to stdio — fail loudly instead. (The linux-arm64
+            // and macOS-x64 prebuilt binaries are core-only.)
+            #[cfg(not(feature = "http"))]
+            if args.iter().any(|a| a == "--http")
+                || std::env::var("DONSETCH_TRANSPORT").as_deref() == Ok("http")
+            {
+                eprintln!(
+                    "mcp: HTTP transport requested (--http / DONSETCH_TRANSPORT=http), but this \
+                     binary was built without the `http` cargo feature. Rebuild with \
+                     --features http, or use a prebuilt binary that includes it."
+                );
+                std::process::exit(1);
+            }
             let _ = std::env::var("DONSETCH_TRANSPORT");
             if args.iter().any(|a| a == "--supervised") {
                 // v3 crash-only design: `--supervised` spawns a child


Mart-Bogdan (approved):
This is nice and correct, but probably would be better if you made it in single PR with #69 

In that case probably readme won't even need gotcha.

imonlinux:
Hey @Mart-Bogdan more than happy to pull the 2 PRs and submit as a single. Just wanting to respect the contribution rules.

Mart-Bogdan:
Hi, @imonlinux, are tour referring to `docs — documentation only` type of checkbox? Or just common curtesy of having atomic commits. 

My reading was that this is when it is documentation **only**, but code change can naturally go alongside with documentation that it fixes.

I believe @dondai44423 would merge both of them anyway.

Only thing that I am bit afraid of is that stray table entry " `mcp --http` silently falls back to stdio" would be outdated.

Mart-Bogdan:
Hey, @imonlinux I have just noticed that your commits that got merged have author set as Claude. 

So your contribution don't count towards your profile. Just leting you know, perhaps you are fine with it.


Also given you are using Claude as well, I wanted to ask, do you have issue with fetch endpoint returning json instead of markdown?

I have resolved it by using local proxy([mcp-text-proxy.js](https://github.com/user-attachments/files/31594968/mcp-text-proxy.js)) and I was planning to make contribution that would resolve it more permanently, but I was blocked from this contribution by my security findigs reagarding URL Handles, which now are fixed by dondai44423 

So I am planning to make it in few days, but I'd like to hear your feedback as fellow Claude user.

PR #78, opened 2026-08-29
Closed

Bnbig:
Closed. Details removed by the author.

PR #79, opened 2026-08-29
feat(search): add parallel explicit query variants

Bnbig:
## Summary

This is a clean resubmission of #78, rebased onto the latest `master` after #77. The implementation is unchanged; test and CLI example data now use generic technical queries.

`web_search` can accept up to two explicit `query_variants` alongside the existing base `query`:

- the single-query path and response shape stay unchanged;
- base + variants run concurrently under one shared deadline;
- each query keeps DonSeTch's existing ranking and returns as a separate result set;
- one global S-handle table covers every successful result in the batch;
- partial failures preserve successful searches, while an all-failed batch remains an honest transient error;
- the MCP schema declares a strict string array with `maxItems: 2`, and the CLI exposes repeatable `--query-variant` flags.

There is deliberately no automatic rewriting or cross-query score fusion. The caller supplies the alternate formulations, and DonSeTch keeps the ranking semantics of each query intact.

```mermaid
flowchart LR
    A[base query + optional variants] --> B[validate and deduplicate]
    B --> C1[existing DonSeek pipeline]
    B --> C2[existing DonSeek pipeline]
    B --> C3[existing DonSeek pipeline]
    C1 --> D[grouped result sets]
    C2 --> D
    C3 --> D
    D --> E[one global S-handle table]
```

## Type

- [x] `feat` — new feature
- [ ] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [ ] `cargo test --features ocr,rerank` passes (upstream x86_64 CI)
- [ ] `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` passes (upstream x86_64 CI)

## Breaking changes

None. Omitting `query_variants` follows the original code path and preserves the existing single-query envelope and cache-key behavior.

## Test plan

Validated locally on Linux ARM64 against current `master` (`dfbfc65`):

- targeted tests for strict schema, CLI parsing, bounds/types, trimming/deduplication, and single-query preservation;
- generated `tools/list` golden fixture;
- `cargo test --locked -- --skip pdf`: 656 unit tests, the RSS soak test, and 4 token-invariant tests passed (matching the upstream ARM64 PDF exclusion);
- `cargo clippy --locked --all-targets -- -Dwarnings -Aunused-imports` passed; the allowance is only for the existing ARM64-only `PathBuf` warning in `src/cpu.rs`;
- `cargo fmt --all -- --check` passed.

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -3276,14 +3276,18 @@ async fn bind_search_handles(
     daemon: &Arc<Daemon>,
     out: &crate::search::SearchOutcome,
 ) -> Vec<String> {
+    let urls: Vec<String> = out.results.iter().map(|r| r.url.clone()).collect();
+    bind_search_urls(daemon, &urls).await
+}
+
+async fn bind_search_urls(daemon: &Arc<Daemon>, urls: &[String]) -> Vec<String> {
     // When handles are disabled (DONSETCH_URL_HANDLES=off), return
     // empty vec — search results show raw URLs instead.
     if !crate::handles::handles_enabled() {
         return Vec::new();
     }
-    let urls: Vec<String> = out.results.iter().map(|r| r.url.clone()).collect();
     let mut ht = daemon.handles.lock().await;
-    let hs = ht.set_search_results(&urls);
+    let hs = ht.set_search_results(urls);
     // Search handles are in-memory only — no flush.
     hs
 }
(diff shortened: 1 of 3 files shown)


Bnbig:
For context, I was comparing DonSeTch with other web-search tools and noticed that trying a few different query formulations could surface relevant evidence that DonSeTch was missing with a single query. So I figured the gap might be partly in query orchestration, not necessarily the ranker itself.
I told Codex to test a small, explicit set of query variants locally and see if it actually helped with the problematic search, then I went to dinner. When I came back, it had seen that it worked and got a bit overexcited and pushed it on its own 😅

If you like the direction, I’m happy to keep working from it. I have a few other ideas too, but I’d rather test and benchmark them one by one first.

Another design idea I have is actually to simplify the tool surface a lot and potentially collapse `web_search`, `web_fetch`, and `web_crawl` into a single tool. I think most of that routing could be handled internally, and it would drastically reduce the token/tool-schema surface the model has to deal with. I’m actively testing this right now with a few benchmarks, and so far it seems to perform basically the same, so I think it could be worth exploring.

dondai44423:
Merged, thanks. The single-query envelope staying byte-for-byte was the deciding detail, and the partial-failure rendering is exactly how the rest of the tool treats honest degradation. Nice work.

Issue #84, opened 2026-08-30
Playwright Chromium auto-discovery misses the chrome-linux64 layout (Playwright >=1.57) — doctor reports "Chrome/Chromium not found"


hobbesd:
## Summary

The Playwright-cache auto-discovery in `known_chrome_paths()` only probes `chrome-linux/chrome`. Playwright changed its Chromium layout to `chrome-linux64/chrome` (Chromium revision ~1200+, Playwright ≈1.57). On those installs the probe silently misses a perfectly usable browser:

- `donsetch doctor` reports `✗ Chrome/Chromium  not found`
- every tier-2 escalation fails with `ghost: no chromium/chrome binary found (set DONGHOST_CHROME)`

…even though Chromium is installed, current, and launches fine.

The comment right above the code states the intent — "Auto-discover it so `donsetch doctor` just works" — so this looks like an unnoticed upstream layout change rather than a deliberate limitation.

## Environment

- DonSeTch 3.4.3 (prebuilt `donsetch-linux-x64`); the same code is on `master` @ `1fe3d4e`
- Linux x86_64, Debian 13 container, glibc 2.41
- Playwright cache: `$HOME/.cache/ms-playwright/chromium-1223/chrome-linux64/chrome`, Chromium 148.0.0.0
- `HOME` is set correctly; the browser is not on `PATH` (normal for Playwright installs, so the `PATH` fallback does not save it)

## Reproduce

With a Playwright ≥1.57 Chromium installed and no system Chrome:

```
$ env -u DONGHOST_CHROME donsetch doctor
  ✗ Chrome/Chromium            not found
    Install Chrome/Chromium, or set DONGHOST_CHROME to a browser path
```

## Evidence that the directory name is the sole cause

Adding a symlink for the old layout and removing it again flips the result deterministically:

```
$ ln -sfn .../chromium-1223/chrome-linux64 .../chromium-1223/chrome-linux
$ env -u DONGHOST_CHROME donsetch doctor
  ✓ Chrome/Chromium            148.0.0.0 (probed) at .../chromium-1223/chrome-linux/chrome

$ rm .../chromium-1223/chrome-linux
$ env -u DONGHOST_CHROME donsetch doctor
  ✗ Chrome/Chromium            not found
```

## Root cause

`src/ghost/mod.rs:168` (v3.4.3):

```rust
if name.starts_with("chromium-") {
    let candidate = entry.path().join("chrome-linux/chrome");
```

## Suggested fix

[...]

Issue #86, opened 2026-08-30
crawl focus scoring is documented as BM25 but is flat token overlap


Mart-Bogdan:
`src/crawl/score.rs` opens with:

 > Frontier relevance scoring — BM25-lite over anchor text + URL path tokens.

The implementation (`score_candidate`, lines 32-45) is not BM25 or a reduction of it. Per query token: `+3.0` if it appears in the anchor, `+1.5` if it appears in the path, each saturating at its first hit; divided by query length; plus a depth prior. There is no IDF, no term frequency, and no document-length normalization — the three things that make BM25 what it is.

This matters because the same repo uses the name correctly elsewhere: `src/extract/focus.rs` is genuine Okapi BM25 (`K1=1.2`, `B=0.75`, real IDF). A reader who knows the fetch path reasonably assumes the crawl frontier ranks the same way, and it doesn't.

The claim had already propagated outward: the `web_crawl` `focus` tool description said "Uses hybrid keyword + semantic matching", which was wrong twice over — the frontier does no semantic matching at all, and its keyword matching isn't BM25. (Corrected separately.)

Two ways to resolve, and which one is right is a maintainer call:

1. **Fix the comment** — call it token-overlap scoring. Cheap, honest, no behavior change.
2. **Fix the code** — implement real IDF over the anchor/path corpus, so a common word like "docs" stops scoring as high as a distinctive one. Better ranking on large sites, but a real behavior change to the crawl frontier.

dondai44423:
Fixed at the code level, option 2. The frontier now has real Okapi IDF: the map phase builds document frequencies from the site's own sitemap inventory and scoring multiplies each query token's hit by ln(1 + (N - df + 0.5)/(df + 0.5)). A token that appears in every /docs URL shrinks; a distinctive one multiplies. No inventory (BFS/content mode, resume) keeps the exact pre-IDF flat weights, and a regression test pins that legacy path at 4.5 plus another proving a rare-token page outranks docs furniture. Tool descriptions (web_fetch and web_crawl focus) now match the code in both places, and the score.rs header is accurate again. Shipped in v3.4.4.

PR #87, opened 2026-08-30
feat(mcp): instructions blurb at initialize, honest focus docs

Mart-Bogdan:
_Note:_ not rebased on master. Master's Windows leg is red on `ghost::sandbox_tests::playwright_discovers_chrome_linux64_layout` (from e670254, issue #84) — the test builds a `chrome-linux64/chrome` fixture, but `playwright_entry_suffixes()` returns only the `chrome-win64/chrome.exe` paths on Windows, so it can only pass on Linux.

As a result this branch conflicts with master in `CHANGELOG.md` only (both sides appended to `[Unreleased]`), which is why GitHub is holding the checks. Left for the hand-merge — say the word and I'll rebase instead. Worth noting a rebase wouldn't make CI green either: `pull_request` builds the merge with master, so that ghost test runs regardless.

**Edit:** Blocked by merge of #89 

## Summary

MCP `instructions` reach the agent's context unconditionally, alongside the system prompt. On harnesses with deferred tool loading — only tool names up front, schemas fetched on demand — this is the only thing that tells an agent these tools exist. Generated from the spec table, so a new tool announces itself with no prose edit.

Kept to ~100 tokens: it is resident in every session whether or not the server is ever used. A new invariant caps it at 150 — the first gate in token_invariants.rs on what we send at handshake time rather than on what a response costs. A golden fixture pins the whole initialize result, holding the package version as a sentinel so a release bump never re-blesses it.

**Fix:**
Both focus descriptions also claimed "hybrid keyword + semantic matching" flatly. On fetch the cross-encoder pass runs only on pages of <=80 blocks and only when the rerank model is already cached from prior search use — plain BM25 otherwise, never a download mid-fetch. On crawl there is no semantic pass at all: frontier candidates score by token overlap on link text and URL path, and a link sharing no token with the query is never enqueued — a hard gate that decides what the crawl reaches and went unmentioned. An agent phrasing a semantic-style query for a BM25 matcher gets worse results, so this is a contract defect rather than wording.

## Type

- [x] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo test --features ocr,rerank` passes (719 lib + 5 integration tests)

## Breaking changes

None at the protocol level — `instructions` is an optional field in the initialize result, and no existing field changed.

Behavior change worth knowing: every session now carries ~100 extra tokens of context whether or not the server is used, and clients that surface `instructions` will display the blurb. On the upside, global/system-prompt nudges that told agents to prefer donsetch can now be dropped — the server announces itself.

## Test plan

Verify performance under real agents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **MCP `instructions` at initialize:** the handshake now carries a
+  short server blurb — one line per tool, generated from the spec
+  table — so a client injects it into the agent's context alongside
+  the system prompt. This is what an agent sees on harnesses with
+  deferred tool loading, which show only tool NAMES up front and
+  fetch schemas on demand: without it, nothing tells the agent
+  these tools exist or when to reach for one. Kept deliberately
+  small (~100 tokens) since it is resident in every session whether
+  or not the server is used, and gated at 150 tokens by a new
+  invariant in `tests/token_invariants.rs` — the first gate in that
+  file on what we send at handshake time rather than on what a
+  response costs. A golden fixture (`tests/fixtures/initialize.json`)
+  pins the whole initialize result; the package version is held as
+  a sentinel there so a release bump never has to re-bless it.
+
 ### Fixed
+
+- **`focus` was described as doing more than it does:** both tool
+  descriptions claimed "hybrid keyword + semantic matching"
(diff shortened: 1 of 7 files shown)


dondai44423:
Merged into master for v3.4.4, thank you. The initialize feature landed exactly as written (instructions blurb, 150-token invariant, golden fixture). The two focus-docstring edits were adapted at merge: the crawl frontier got real IDF scoring today (issue #86), so the final wording describes the IDF version rather than the flat-overlap version your PR described. Credited in the merge commit.

PR #82, opened 2026-08-30
feat(search): add SerpApi as a BYOK search provider

mnaza:
## Summary

Adds [SerpApi](https://serpapi.com) as a new BYOK search provider (`donsetch keys add serpapi <key>`), alongside the existing Serper.dev provider. Follows the exact same adapter pattern as `src/search/byok/serper.rs`: intent-based routing (google_scholar engine for paper queries, tbm=nws for news), and the same key-state error classification (invalid/rate-limited/credit-depleted) as the other providers.

Differs from `serper.rs` in one structural way: SerpApi authenticates via an `api_key` query parameter rather than a header — reqwest's `.query()` percent-encodes it safely.

## Type

- [x] `feat` — new feature

## Checks

- [x] `cargo test --lib` passes (703/703, including 5 new unit tests for the JSON-parsing logic in `serpapi.rs`)
- [x] `cargo clippy --lib -- -Dwarnings` passes (zero warnings)
- [x] `cargo fmt --all -- --check` passes
- [ ] `cargo test --features ocr,rerank` / `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` — not run locally (heavy native/ML deps not set up in my environment); this change only touches `src/search/byok/` and `src/cli/keys.rs`, unrelated to the ocr/rerank features, but I'd appreciate CI confirming the full matrix.
- [ ] CI on all 3 platforms — pending, will watch after opening this PR.

## Breaking changes

None. Purely additive: a new opt-in BYOK provider, registered in `PROVIDERS`/the dispatcher, doesn't touch existing provider behavior.

## Test plan

- Added `parse_results()` as a pure, testable function (mirroring the `parse_key()` convention in `brightdata.rs`, since these HTTP adapters aren't otherwise unit-tested in this codebase) and covered it with 5 unit tests: organic results, news results (different JSON key), entries missing a `link` (dropped), a missing results key (empty output), and a missing `position` field (defaults to score 1.0).
- Confirmed the output `SearchHit` shape is compatible with the existing `to_merged()` dedup/ranking pipeline in `mod.rs` (no changes needed there).
- **Caveat:** I don't have a live SerpApi account, so the HTTP status/error-body classification (particularly the 429 message-sniffing for rate-limit vs. plan-exhaustion, and the "HTTP 200 with an `error` JSON field" fallback path) is based on SerpApi's public docs, not verified against a real account. Flagged in a code comment; happy to adjust once someone can test against a live key.

## Contributor note

[...]

PR #88, opened 2026-08-30
feat(search): add Brave Search API as a BYOK search provider

mnaza:
## Summary

Adds the official, keyed [Brave Search API](https://api.search.brave.com) as a second BYOK search provider (`donsetch keys add bravesearch <key>`), following the same shape as #82 (SerpApi). Named `bravesearch` — deliberately distinct from the existing keyless `brave` SERP scraper in `src/search/engines.rs`, so there's no ambiguity between the two.

Two differences from `serpapi.rs`/`serper.rs` worth calling out:
- Auth is `X-Subscription-Token: <key>` (a header, like Serper — not a query param like SerpApi).
- Brave doesn't return a per-result `position` field, so relevance score is derived from array order instead.
- Web-search results live at `web.results[]`; news-intent queries hit a separate `/res/v1/news/search` endpoint with a top-level `results[]`.

## Type

- [x] `feat` — new feature

## Checks

- [x] `cargo test --lib` passes (713/713, including 4 new unit tests for `bravesearch.rs`'s JSON-parsing logic)
- [ ] `cargo test --features ocr,rerank` / `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` — not run locally (same as #82, heavy native/ML deps not set up here); this change only touches `src/search/byok/` and `src/cli/keys.rs`.

## Breaking changes

None. Purely additive.

## Test plan

- `parse_results()` factored out as a pure, testable function (same convention as `serpapi.rs`/`brightdata.rs`), covered by 4 unit tests: basic ordering/scoring, entries missing `url` (dropped), empty array, and a missing `description` (defaults to empty snippet).
- Verified the `SearchHit` output is compatible with `to_merged()` in `mod.rs` — no changes needed there.
- Cross-checked the endpoint paths, header name, response field names, and `count` max (20) against Brave's public API docs.
- **Caveat**, same as #82: I don't have a live Brave Search API subscription, so the exact wording Brave uses in 429 error bodies (for the message-sniffing that distinguishes rate-limit vs. quota-exhaustion) is inferred from docs, not a real account. Brave's rate-limit docs do confirm there's no 402 status in their API (documented in a comment in the diff), unlike Tavily/SerpApi/Serper/Brightdata.

## Contributor note

Follow-up to #82 (SerpApi), same BYOK-provider pattern. Happy to adjust the error-classification wording once someone can test against a live key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   alongside the existing Serper.dev provider. Routes by intent like
   the other providers: `google_scholar` engine for paper queries,
   `tbm=nws` for news.
-- **Brave Search API BYOK provider:** `donsetch keys add bravesearch
-  <key>` wires up the official, keyed
+- **Brave Search API BYOK provider:** `donsetch keys add bravesearch <key>`
+  wires up the official, keyed
   [Brave Search API](https://api.search.brave.com) — distinct from
   the existing keyless `brave` SERP scraper. Uses a dedicated news
   endpoint for news-intent queries.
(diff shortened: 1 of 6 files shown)


dondai44423:
Merged locally as cafc930 (the PR's base had drifted past an admin squash). Brave's docs confirm the count semantics: web max 20, news max 50, so the shared 1..20 clamp stays inside both. Thank you.

PR #90, opened 2026-08-31
fix(crawl): focus_match compound-term check crosses word boundaries

mnaza:
## Summary

Found while reading through the fresh real-IDF scoring code from #86 (`src/crawl/score.rs`). `focus_match()` — the crawl frontier's hard focus gate — has two ways to match a compound query term (one containing `_` or `-`, e.g. `spawn_blocking`): (a) the full compound form as a substring, or (b) all its fragments as tokens, any order.

Check (a) was a raw `String::contains` against the lowercased but otherwise unmodified path/anchor text. That crosses word boundaries: `auto-complete` is a literal substring of `auto-completed`, even though this codebase's own stemmer (`stem_en`'s `-ed` rule) turns `"completed"` into `"complet"` — a different token from `"complete"` everywhere else matching happens (check (b), `score_candidate`, etc). So `focus_match("", "/features/auto-completed-suggestions", "auto-complete")` returned `true`, letting the crawler enqueue an off-topic page purely because of an accidental substring collision.

Replaces the raw substring check with `contains_subsequence()`: does the compound term's tokenized fragments appear as a contiguous, in-order run within the already-tokenized `path_toks`/`anchor_toks`? Since those went through the same tokenizer/stemmer as everything else, this is boundary-safe and stemming-consistent by construction.

## Type

- [x] `fix` — bug fix

## Checks

- [x] `cargo test --lib` passes (719/719, including a new regression test)

## Breaking changes

None. This only tightens an over-permissive match into a correct one; every case that still matches under the new logic matched under the old logic too (verified: the only paths where old/new logic differ are exactly the word-boundary-crossing false positives being fixed — case (b), which is untouched, keeps working for the "all fragments, any order" path).

## Test plan

- New regression test `focus_match_compound_substring_respects_word_boundaries`: covers the false positive via path text, the same false positive via anchor text, and a true-positive control (`/docs/auto-complete-guide`, where the term appears as its own contiguous tokens) to prove the fix isn't just "always stricter."
- Traced through `focus_match_compound_in_path` and `focus_match_compound_no_false_positive` (existing tests) by hand to confirm they still pass for the right reason under the new subsequence-based logic.
- Reasoned through whether the old substring behavior could ever have caught a legitimate match the new logic now misses — every case where they'd differ is, by construction, a word-boundary-crossing artifact (the bug itself), not a real distinct match worth preserving.

## Contributor note

Not something I went looking for — found it by reading the surrounding code while getting familiar with the new focus-scoring machinery from #86.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to DonSeTch are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **`focus_match` compound-term check crossed word boundaries:** the
+  crawl frontier's hard focus gate matched a compound query term
+  (e.g. `auto-complete`) against any URL/anchor text containing it
+  as a raw substring, so `auto-completed` (a different word once
+  stemming strips `-ed`) falsely counted as a match. The full-form
+  check is now a contiguous token-subsequence match instead of a
+  string `contains`, closing the same word-boundary gap the
+  existing fragment-based check (issue #86 follow-up) already
+  guarded against.
+
 ## [3.4.4] - 2026-08-31
 
 ### Added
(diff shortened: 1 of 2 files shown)


dondai44423:
Merged, shipped on master as a4a8e2c for the next release. Verified two ways: the new regression test fails on the pre-fix base (the false positive really fires) and passes after; and a live crawl of docs.rs tokio with a compound focus still enqueues the spawn_blocking page correctly. The token-subsequence helper is the right shape here, and the v2.3.1 attribution checked out against history (3e62505). Thank you.

PR #96, opened 2026-09-01
cloakbrowser

MasterLegoYoda:
basically, added cloak browser as an alternative to the regular headless chromium. (for real this time)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -48,6 +48,16 @@ docker-compose*.yml
 .dockerignore
 Dockerfile*
 
+# CloakBrowser is a user-managed runtime dependency. Never send downloaded
+# archives or binaries to Docker build context; the image only contains
+# DonSeTch itself and an optional system Chromium installation.
+cloakbrowser/
+**/cloakbrowser-*.tar.gz
+**/cloakbrowser-*.zip
+**/Chromium.app/
+**/chromium-*/chrome
+**/chromium-*/chrome.exe
+
 # Development files
 *.log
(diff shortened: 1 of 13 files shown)


dondai44423:
Merged. Solid work, thank you.

What I verified before merge, against the live CloakHQ release:
- Download chain is end to end real: I fetched the pinned v146 release's SHA256SUMS + sig from GitHub and checked the pinned Ed25519 key against them with openssl. It matches the key CloakHQ ships in their own npm wrapper. Sig verifies clean.
- Then ran the actual installer: download, manifest check, sha256, extract, version probe, and a doctor --deep launch on the cloak backend. 15/15, healthy. It is a true drop-in for ghost, exactly as designed. Did not lay a finger on the tier-1 fetch path.

Three things changed in the follow-up on master, per the strict opt-in rule:
1. A bare CLOAKBROWSER_BINARY_PATH no longer switches the backend under auto. Someone else setting that var for another tool would have silently selected cloak, which violates the 'never used unless selected' rule. Selection now requires the explicit backend value. Regression test added.
2. Empty backend value = auto (safe for Docker-style empty defaults).
3. Status line + docstring wording match the actual semantics.

One honest caveat in the README/docs: the free v146 build ages as their detection arms race moves on. Users on it can bump CLOAKBROWSER_VERSION or point CLOAKBROWSER_BINARY_PATH at a newer build they obtained directly, without any changes on our side. That flexibility is exactly why the binary path override is first class here.

PR #97, opened 2026-09-01
fix(ghost): Windows profile winlock can be stolen from a still-live daemon

mnaza:
## Summary

Found by reading through the Windows profile-lockfile logic added in a recent commit (the flock-equivalent for two daemons sharing the same Chromium profile dir).

**The bug**: the lockfile's mtime is set once at creation and never touched again. Staleness is judged purely by that mtime being >10 minutes old (`WINLOCK_STALE_AFTER`). But `GhostGuard::drop` kills the Ghost unconditionally on every guard drop on Windows (`manager.rs`) — so the lock is normally held for exactly one call's duration. The exception: a single `actions` call can legitimately run past 10 minutes in one unbroken hold (up to `MAX_STEPS=16` steps, each `wait_selector`/`wait_text` capped at 60s — `actions.rs`). If a second daemon starts up during that window, it sees the un-refreshed creation-time mtime, concludes the still-live holder is dead, deletes its lockfile, and steals the shared profile — the exact two-daemon collision this lock exists to prevent.

**The fix**: a background task refreshes the lockfile's mtime every 2 minutes (`WINLOCK_HEARTBEAT`) for as long as a Ghost holds it, comfortably below the 10-minute staleness window, aborted in `Drop` alongside the existing `fetch_guard` task (same pattern). Both the initial lock and the heartbeat's periodic reopen now request `FILE_SHARE_DELETE` explicitly, so a heartbeat tick can never block `Drop`'s cleanup on exit.

## Type

- [x] `fix` — bug fix

## Checks

- [x] `cargo test --lib` — 727/729 (2 unrelated, pre-existing failures in `fetch::guards::tests::{ensure_url_safe_blocks_literal,ensure_url_safe_fail_closed_on_dns_error}` — these do live DNS/loopback checks and fail in my sandboxed dev environment's network setup; unrelated file, not touched by this diff, confirmed via `git show --stat`)
- [x] `cargo test --lib ghost::` — 52/52 (the new Windows-gated test correctly compiles out on Linux and doesn't run here)
- [x] `cargo clippy --lib -- -Dwarnings` — zero warnings
- [x] `cargo fmt --all -- --check` — clean
- [ ] CI on all 3 platforms — pending, especially interested in the Windows leg here.

## Breaking changes

None.

## Test plan — and an important caveat

I do not have access to a Windows machine, and a full cross-compile from this Linux box fails early at BoringSSL's native-assembly build stage (same limitation as my earlier `known_chrome_paths` macOS fix), so **I could not execute or reproduce the actual race on real Windows.** This fix rests on static reasoning about `FREEZE_AFTER`/`REAP_AFTER`/`GhostGuard::drop`'s interaction (traced through `manager.rs` and `actions.rs` by hand) plus a from-scratch independent code review pass, not a live repro.

What I *could* verify:
- A new `#[cfg(windows)]` regression test pins the safety-margin invariant between `WINLOCK_HEARTBEAT` and `WINLOCK_STALE_AFTER` (3x margin) — it will compile-check and run on the Windows CI leg, but obviously can't simulate the actual two-process race.
- Traced every `Ghost`-dropping code path (`acquire`'s relaunch, the reaper, shutdown) to confirm none of them leak a `Ghost` without running `Drop` (no `mem::forget`/`Box::leak`/detached-Arc patterns anywhere in the codebase for this type).
- Confirmed `tokio::spawn` is valid at this call site (the pre-existing `fetch_guard` task is spawned the same way, a few lines earlier in the same constructor).

What CI's Windows leg will **not** catch, and where I'd specifically want a human with real Windows access to look before merging:
- The actual two-daemon race end-to-end (one daemon busy past 10 minutes via a long `actions` script, a second daemon starting mid-way, confirming it backs off instead of stealing the profile).
- Whether `File::set_modified()` reliably and promptly updates the visible mtime under real-world conditions (AV/EDR software transiently locking the file, non-NTFS volumes, etc.) — heartbeat write failures are currently silent (swallowed via `if let Ok(...)`), so a systemic failure would only surface as an eventual, hard-to-diagnose collision with no diagnostic trail. Happy to add `DONGHOST_DEBUG` logging on heartbeat failures in a follow-up if that'd help.

## Contributor note

This went through an unusually adversarial self-review round given the "can't test on Windows" risk — the first draft's doc comments incorrectly blamed a "continuously-busy Ghost" (a framing that doesn't actually apply on Windows, since the Ghost dies after every single guard use there) instead of the real trigger (one long `actions` call), and the review also caught a plausible Drop-vs-heartbeat file-sharing race that the `FILE_SHARE_DELETE` addition closes. Both are fixed in the second commit on this branch. I'd rather have this land after a human with Windows access has poked at it than assume it's correct — genuinely open to "hold this until someone can verify on real hardware" if that's the right call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   DevTools line. A create_new profile lockfile now mirrors the
   unix flock: the loser diverges to a temp profile, and a stale
   lock left by a dead daemon recovers by age (10 min).
+- **Windows profile lockfile could be stolen out from under a live
+  daemon:** the lockfile's mtime was never refreshed after
+  creation, so a continuously-busy Ghost (which only freezes after
+  20s idle and only reaps after 10min *frozen* — it can run far
+  longer than that without ever hitting either) could hold the
+  lock well past the 10-minute staleness window. A second daemon
+  starting mid-way would see the un-refreshed mtime, mistake the
+  still-live holder for abandoned, and steal the profile — the
+  exact collision the lock exists to prevent. The lockfile's mtime
+  now gets refreshed every 2 minutes for as long as a Ghost holds
+  it.
 - **Browser fingerprint noise:** the ghost no longer runs Chrome
   default apps or extensions, killing the surprise-component
   detection class (enumerable extensions, default-app traffic)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,15 +56,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   lock left by a dead daemon recovers by age (10 min).
 - **Windows profile lockfile could be stolen out from under a live
   daemon:** the lockfile's mtime was never refreshed after
-  creation, so a continuously-busy Ghost (which only freezes after
-  20s idle and only reaps after 10min *frozen* — it can run far
-  longer than that without ever hitting either) could hold the
-  lock well past the 10-minute staleness window. A second daemon
-  starting mid-way would see the un-refreshed mtime, mistake the
-  still-live holder for abandoned, and steal the profile — the
-  exact collision the lock exists to prevent. The lockfile's mtime
-  now gets refreshed every 2 minutes for as long as a Ghost holds
-  it.
+  creation. Windows kills the Ghost on every guard drop (unlike
+  Linux/Xvfb's warm-freeze path), so the lock is normally held for
+  one call's duration — but a single long `actions` script (up to
+  16 steps, each wait capped at 60s) can legitimately outlast the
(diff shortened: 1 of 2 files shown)


dondai44423:
Merged. This was a correct catch of a real defect: the lock's mtime was a birth timestamp and nothing refreshed it, so any ghost held past 10 minutes looked stale and could be deleted out from under a live daemon. The heartbeat every 120s with the 600s stale window keeps three heartbeats of margin, and the Drop ordering (abort then remove) matches the fetch_guard pattern.

Verification: 766 tests locally including the new winlock invariant test, full-feature clippy and fmt clean. The Windows leg cannot execute here, so the merge relies on your invariant test plus the ability of the heartbeat task to keep the mtime ahead of the stale window; the release CI will exercise the build on the Windows runner. The changelog wording was tightened in the follow-up for consistency.

PR #103, opened 2026-09-01
fix(search): ignore tracking parameters during URL deduplication

adaaaaaaaaaaaaaaaaaaaaaa:
### Summary

Search consensus treated one page as multiple results when providers attached
analytics parameters. I remove only known tracking keys from the internal
deduplication key while preserving meaningful query bytes, ordering, repeated
parameters, and encoded delimiters.

This improves duplicate collapse and cross-provider agreement without claiming
that arbitrary query strings are interchangeable.

### Verification

- Covered ordinary, case-insensitive, and percent-encoded tracking keys.
- Preserved an unclassified `ref` parameter and meaningful page values.
- Proved repeated-key order and encoded delimiters cannot collide.
- Full v3.5.0 validation is green.

### Alternatives considered

Sorting or decoding the full query was rejected because query order and encoded
syntax can be application semantics. A large generic ignore list was rejected
because false equivalence is worse than failing open on an unknown tracker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   downstream shell heuristics could mark valid, compact PDF output as thin
   and suggest browser rendering. Documents already identified through PDF
   page metadata now bypass HTML-only shell classification.
+- **Tracking parameters split identical search results during URL
+  deduplication (PR #103):** normalized URL keys retained analytics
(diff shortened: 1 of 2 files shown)


Mart-Bogdan:
@adaaaaaaaaaaaaaaaaaaaaaa please provide changelog entry under `## [Unreleased]` heading

adaaaaaaaaaaaaaaaaaaaaaa:
Added the `Unreleased` changelog entry in 02e92fa and rebased the branch onto the current master. The tracking-parameter deduplication regression test still passes. Thanks.

PR #108, opened 2026-09-02
fix(pdf): bookmark titles get 2 trailing NULs from a byte/unit mismatch

mnaza:
## Summary

Found during a broader security-focused read-through of the PDF/PDFium FFI layer (nothing memory-unsafe found there overall — this is a text-correctness bug, not a security issue).

`FPDF_GetMetaText`/`FPDFBookmark_GetTitle` report their length in **bytes** (including the UTF-16LE NUL terminator — see `fpdf_doc.h`), but both `get_meta()` and `bookmark_title()` in `src/pdf/engine.rs` passed that byte count straight into `decode_utf16(buf, len_units)`, which expects a UTF-16 **unit** count. The mismatch reads past the real decoded string into the buffer's zero-initialized slack (both call sites deliberately over-allocate to `n/2 + 2` units for a safe refetch — that slack is exactly what leaks through).

Practical impact differs between the two call sites:
- `get_meta()`'s output is unaffected — its caller already does `s.chars().filter(|&c| c != '\0').collect()` before returning, silently stripping the garbage tail.
- `bookmark_title()`'s result flows straight into `OutlineItem.title` with nothing downstream to filter it: **every extracted PDF bookmark/outline title carries exactly 2 trailing NUL characters.**

Fix: added `decode_utf16_from_byte_count()`, the correct bytes→units conversion, mirroring the sibling `field_string()` helper in `src/pdf/forms.rs` (used for AcroForm field values), which already divides by 2 correctly. Switched both call sites to it.

## Type

- [x] `fix` — bug fix

## Checks

- [x] `cargo test --lib pdf::engine::` passes (4/4 new unit tests)
- [x] `cargo test --lib` passes (744/744)

## Breaking changes

None.

## Test plan

- 4 new unit tests directly exercise `decode_utf16_from_byte_count()`'s byte→unit math: strips the terminator and padding correctly, handles an empty title, never reads past the buffer even with a wildly-wrong byte count (defensive against a hostile/corrupt PDF), and agrees with calling `decode_utf16()` directly using the hand-computed correct unit count.
- Honest caveat: these test the new helper directly, not the old buggy call sites end-to-end (`bookmark_title`/`get_meta` need a real PDFium document/bookmark handle, and there's no existing fixture-driven outline-extraction test in the codebase to hook into for this fix's scope). The byte/unit math is the actual bug and is what's pinned.
- Grepped for other `decode_utf16` call sites — these are the only two, and they're now both fixed; `field_string` in `forms.rs` (a structurally similar helper) already did the division correctly, confirming the bug was isolated to these two spots.

## Contributor note

Not the primary target of the security sweep this came out of — this is the one non-security (text-correctness) finding from that pass. Happy to also look at filing a follow-up for `get_meta`'s post-resize decode always using the first call's byte count (`n`) rather than the refetch's (`n2`) — almost certainly benign since PDFium's contract should make them equal once the buffer is sized correctly, but noting it in case it's worth a defensive tweak; didn't want to bundle an unrelated speculative change into this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   result cards can wrap their breadcrumb and heading in one outer link, so
   reading the whole anchor produced noisy titles. Yahoo parsing now extracts
   the dedicated heading first and keeps the outer text as a fallback.
+- **PDF bookmark titles carried two trailing NUL characters:**
+  `FPDF_GetMetaText`/`FPDFBookmark_GetTitle` report their length in bytes
+  (including the UTF-16 NUL terminator), but the decode call was treating
+  that count as UTF-16 units — a mismatch that read past the real string
+  into the buffer's zero-initialized slack. `get_meta`'s output was
+  unaffected (it already strips all `'\0'` chars), but every extracted
+  outline/bookmark title picked up two invisible trailing NULs. Both call
+  sites now convert bytes to units first, matching the sibling
+  `field_string` helper in `forms.rs`, which already did this correctly.
 
 
 ## [3.5.0] - 2026-09-01
--- a/src/pdf/engine.rs
+++ b/src/pdf/engine.rs
@@ -399,6 +399,19 @@ fn decode_utf16(buf: &[u16], len_units: usize) -> String {
     String::from_utf16_lossy(&buf[..end])
 }
 
+/// Decode a UTF-16LE buffer using a PDFium-reported BYTE count, as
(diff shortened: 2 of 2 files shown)


dondai44423:
Merged. Confirmed against the PDFium headers: FPDF_GetMetaText and FPDFBookmark_GetTitle both report bytes (including the UTF-16 NUL terminator), and both call sites fed that byte count into a unit-count decode, so the two trailing NULs from the over-allocated slack leaked into every bookmark title. The bytes-to-units helper fixes both, and the defensive clamp against a corrupt PDF reporting an insane size is the right call for hostile input.

PR #109, opened 2026-09-02
feat(byok): add SerpBase (Google SERP) as a BYOK provider

### LINKED ISSUE #94: feat(search): add SerpBase as a BYOK search provider
Hi — DonSeTch has been my go-to for agentic web research since the v3 release; the keyless search plus `fetch S3` warm-handoff is the combo I use most. I've also been following the BYOK provider work in #82/#88 — the SerpApi and Brave adapters landed fast and the key-rotation/cooldown design is a nice touch.

One thing I keep running into: for plain Google SERP results the keyless engines work, but on sites that serve the full JS-heavy page, or when I need richer Google verticals (news, shopping, AI Overviews) with a guarantee of a clean structured response, I end up configuring an external provider anyway. The current BYOK list is solid, but the Google-specific options are either per-result expensive or missing the rich-result fields entirely.

Would you consider adding SerpBase (https://serpbase.dev) as another BYOK provider? It's a Google Search Results API that returns structured JSON — organic results with positions, AI Overviews, featured snippets, and rich results — via a single `?q=&num=` GET. It's pay-as-you-go with a free tier, and there's no subscription. The response shape maps almost 1:1 to what `serpapi.rs` already parses (`organic_results`, `news_results`), so the adapter would follow the exact same layout discipline the maintainer called out in #82.

Why it fits:
- The BYOK pattern is already established (SerpApi, Serper, Brave, Exa, Tavily, TinyFish, Parallel, Bright Data); this is just one more adapter behind `donsetch keys`, same `KeyError` classification and key-state handling.
- The `Intent` routing already special-cases Google (`google_scholar` for papers, `tbm=nws` for news) — SerpBase's Google verticals would slot into that routing directly.
- For users who want a lightweight, no-scraping-maintenance fallback for the Google surface specifically, this gives them a pay-as-you-go option at a lower entry cost than the existing Google-facing providers.

Happy to draft the adapter (`serpbase.rs` mirroring `serpapi.rs`, with tests) if there's interest — I've got the pattern down from reading the existing providers. No pressure either way, and thanks for keeping the keyless experience the default.

gefsikatsinelou:
## Summary
Adds SerpBase (https://serpbase.dev) as another BYOK Google SERP provider, per the discussion on #94. It mirrors the existing `serpapi`/`serper` adapters exactly — one adapter file plus registry/help/README lines.

## Changes
- **New**: `src/search/byok/serpbase.rs` — `search()` adapter + pure `parse_results()` (4 unit tests)
- **Updated**: `src/search/byok/mod.rs` — module + dispatch arm
- **Updated**: `src/search/byok/store.rs` — provider registry + list help
- **Updated**: `src/cli/keys.rs` — `keys add` help line
- **Updated**: `README.md` — `keys add serpbase` example

## API contract notes (verified against serpbase.dev docs)
- `POST https://api.serpbase.dev/google/search`, JSON body `{q, num, hl, gl}`, auth via `X-API-Key` header
- Business errors can come back as HTTP 200 with a non-zero `status` envelope (`1001` = unauthorized → mapped to `KeyError::InvalidKey` so rotation marks the key dead; others → `UnknownError`)
- HTTP 401/403 → `InvalidKey`, 402 → `CreditDepleted`, 429 → `RateLimited`, 5xx → `ServerError`
- Results under `organic[]` (`position`/`title`/`link`/`snippet`) → mapped to `SearchHit` with position-derived relevance score, same as serpapi/serper
- `num` capped at 10 to match the other providers
- No new dependencies, no behavior change when the key is absent

## Testing
- 4 new unit tests on `parse_results` (organic mapping, drops entries without link, empty on missing key, position default)
- Full suite: 733 passed, 0 failed
- `cargo fmt --check` and `cargo clippy --lib` clean

Closes #94 (feature request).

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,6 +472,7 @@ donsetch keys add tavily tvly-...       # Tavily
 donsetch keys add exa sk-exa-...        # Exa (stackable)
 donsetch keys add serper ...            # Serper.dev
 donsetch keys add serpapi ...           # SerpApi
+donsetch keys add serpbase sb-...       # SerpBase Google SERP (100 free searches)
 donsetch keys add bravesearch ...       # Brave Search API
 donsetch keys add tinyfish sk-...       # TinyFish (free tier)
 donsetch keys add parallel nKil3...     # Parallel AI (fast mode)
--- a/src/cli/keys.rs
+++ b/src/cli/keys.rs
@@ -9,7 +9,7 @@
 //!   donsetch keys import <path>            Import keys from a file
 //!   donsetch keys clear                     Remove all keys
 //!
-//! Providers: tavily, exa, serper, serpapi, bravesearch, tinyfish, parallel, brightdata, unlocker
+//! Providers: tavily, exa, serper, serpapi, serpbase, bravesearch, tinyfish, parallel, brightdata, unlocker
 
 use super::{bold, dim, green, red};
 
@@ -466,6 +466,7 @@ fn print_help() {
     println!("    exa        Exa AI Search (api.exa.ai)");
     println!("    serper     Serper.dev Google SERP (google.serper.dev)");
     println!("    serpapi    SerpApi Google SERP (serpapi.com)");
+    println!("    serpbase   SerpBase Google SERP (serpbase.dev)");
     println!("    bravesearch Brave Search API (api.search.brave.com)");
     println!("    tinyfish   TinyFish Search (api.search.tinyfish.ai)");
     println!("    parallel   Parallel AI Search (api.parallel.ai) : fast mode");
--- a/src/search/byok/mod.rs
+++ b/src/search/byok/mod.rs
@@ -19,6 +19,7 @@ mod brightdata;
 mod exa;
 mod parallel;
 mod serpapi;
+mod serpbase;
 mod serper;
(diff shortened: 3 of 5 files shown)


dondai44423:
Merged. Your adapter checked out against the live serpbase.dev docs on all the points that matter: POST + X-API-Key + organic array + the 1001 unauthorized envelope mapping to InvalidKey so rotation retires dead keys. Nice catch on the outdated GET/organic_results contract too.

Two follow-ups landed on master after the merge: the docs list q/hl/gl/page/device as the request parameters, no num, so the body no longer sends one (results are truncated to the caller's max after parsing instead), and the [Unreleased] changelog entry is in (closes #94). README stays exactly as you wrote it.

gefsikatsinelou:
Thanks for the thorough check against the live docs @dondai44423 — glad the adapter held up on all the points that matter. Makes sense on the follow-ups: dropping `num` from the body (results truncated to the caller's max after parsing) and the [Unreleased] changelog entry closing #94. Appreciate you keeping the README as written, and great to hear the plugin-style BYOK idea from @Mart-Bogdan is shipping next — happy to help test that too when it's ready.

Issue #95, opened 2026-09-01
ghost: concurrent Xvfb :99 startup emits misleading failure and degrades escalation


fwhskr:
## Summary

Donsetch 3.4.4's ghost browser escalation reports an Xvfb start failure when concurrent sessions race for the shared `:99` display (or when its lock/socket is stale). The binary is installed and can fall back to headful off-screen mode, but the diagnostic still advises installing Xvfb. The Pi package extension forwards all MCP worker stderr to the Pi process stderr, so this warning becomes a user-facing TUI popup.

## Evidence

In a disposable Debian container with Xvfb and Chromium installed, a private fake `Xvfb` executable that exits non-zero was placed first in `PATH`. No host display, lock, socket, or worker was touched. A direct `web_fetch` MCP call with `tier: "2"` produced:

```text
[ghost] Xvfb start failed: ghost: Xvfb failed to start (install: apt install xvfb / pacman -S xorg-server-xvfb), falling back to headful off-screen mode
```

With the normal Xvfb executable, `donsetch doctor` reports Xvfb available and the display is reused. The issue is therefore the ghost tier's shared-display startup coordination and misleading fallback diagnostic, not a missing package.

## Requested fix

Please make the Xvfb helper safely probe/reuse an already-started `:99`, serialize startup, and distinguish a stale lock from a missing executable. The diagnostic should not suggest installing Xvfb when the executable is present; it should include the actual launch/lock error and preserve a successful browser escalation after another session owns the display.

Mart-Bogdan:
Theoretically 99 can also be used by another user's process as well.

Is it possible for Xvfb  to probe different ids and just return it to us.

dondai44423:
Confirmed and fixed on master, shipping in v3.5.0 later today.

The race was real: concurrent starters could both run the stale-cleanup pkill on :99, killing whichever had a live Xvfb up first, then degrade to headful off-screen while blaming the package manager. And the extension was raw-forwarding daemon stderr into pi's TUI, so the misleading warning surfaced as a popup.

Both fixed:

- Xvfb startup is serialized through a create_new gate (per-display, /tmp/.donsetch-xvfb-N.lock). One coordinator starts, everyone else blocks inside the gate and reuses the display once it is up. Escalation no longer degrades when another session owns :99. Stale gates recover after 30s.
[...]

PR #111, opened 2026-09-02
fix(pi-extension): gate startup banner behind DONSETCH_DEBUG

OwaisQuadri:
The `[donsetch] N tools registered` banner in `npm/pi-extension.ts` printed unconditionally on every `session_start` via a raw `process.stderr.write`, which bypasses pi's TUI paint cycle. That corrupted the terminal viewport on every normal run, and got worse under parallel background agents, where each process's banner interleaved with the others on the same screen (garbled repeated lines above the status bar).

Gates it behind `DONSETCH_DEBUG`/`DEBUG`, matching the existing gate already applied to the daemon's forwarded stderr diagnostics (issue #95). A normal session now prints nothing from this path; debug mode still shows it.

CHANGELOG entry included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **pi-extension: startup banner corrupted the viewport:** the
+  `[donsetch] N tools registered: ...` line printed on every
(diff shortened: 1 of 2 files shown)


dondai44423:
Merged. Right fix: that banner was the only session_start path still holding a raw process.stderr.write, and it interleaved badly across parallel agents. Hoisting the DEBUG gate to module scope and reusing it for both the banner and the daemon stderr forwarding matches the #95 direction.

PR #129, opened 2026-09-05
refactor(cli): Route CLI titles through one display-name constant

Mart-Bogdan:
# Route CLI titles through one display-name constant
## Summary

The product name `DonSeTch` was hardcoded in 16 places across the CLI. It now
comes from one constant, `donsetch::DISPLAY_NAME` in `src/display_name.rs`,
which the exe's `FileDescription` and the MCP `serverInfo.title` already read.
Rendered output is unchanged: every substitution expands to the same string.

Stacked on the Windows version resource PR, which introduced the constant.
Base that branch, not master.

## What moved

| Site | Count |
|---|---|
| `proxy` subcommand titles | 9 |
| `doctor`, `update`, `rollback`, `status`, `version` titles | 5 |
| `keys` help prose | 2 |

`USAGE:`/`Usage:` lines are deliberately untouched — those show the command as
typed, which is the lowercase `donsetch`, not the product name.

## Why

A rename is one edit instead of a grep, and the CLI can no longer drift from
the two places the name is already consumed: the Windows `FileDescription`
stamped by `build.rs`, and `serverInfo.title` in the MCP handshake. That drift
is the failure this prevents — three surfaces disagreeing about what the
product is called, with nothing to catch it.

## Type

- [ ] `feat` — new feature
- [ ] `fix` — bug fix
- [x] `refactor` — no behavior change
- [ ] `test` — tests only

## Checks

- [x] `cargo test --features ocr,rerank -- --test-threads=1` passes\
      (serial run needed: `tests/auth_login.rs` shares one state dir across the
      whole binary, so two logout tests race under parallel execution — pre-existing,
      untouched by this PR)

## Breaking changes

None. Output is byte-identical — the constant expands to the same literal each
call site held before.

## Test plan

- Rendered a title through the changed path to confirm the output is unchanged:

```
DonSeTch Proxy Configuration
─────────────────────────────────────────────────────────

  No proxies configured.
  Use `donsetch proxy add <url>` to add one.

─────────────────────────────────────────────────────────
```

- `grep -rn "DonSeTch" src/` returns no string literals outside
  `src/display_name.rs`; the remaining hits are doc comments and prose.
- Full suite and clippy as above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   since 2025-06-18), so clients that prefer it now show `DonSeTch`
   instead of the `donsetch` package id. Older clients ignore the field.
 
+### Changed
+
+- **One source for the product name:** hardcoded in 16 places across the
+  CLI, it now lives in `src/display_name.rs` (`donsetch::DISPLAY_NAME`),
+  read by the CLI titles, the MCP title and build.rs alike : the last by
+  `include!`, since a build script cannot `use` items from the crate it
+  builds.
+
 ### Fixed
 
 - **False-like private-egress values no longer disable SSRF guards:**
--- a/src/cli/doctor.rs
+++ b/src/cli/doctor.rs
@@ -14,6 +14,7 @@
 
 use std::path::Path;
 
+use crate::DISPLAY_NAME;
 use crate::cli;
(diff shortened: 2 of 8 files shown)


Mart-Bogdan:
I have rebased it on top of newest `master`
`just test`:
```
Summary [  27.935s] 899 tests run: 899 passed, 0 skipped 
```

PR #168, opened 2026-09-09
fix(ghost): pool spills hosts into free slots instead of funneling everything into slot 0

mnaza:
## The bug (e8824d2, warm browser pool)

The daemon runs **one** browser profile, so every ghost acquire computes the same persona key. `pick_slot` ranked *"any warm same-persona slot"* above free slots — so after the first launch, every job to every host resolves to slot 0, forever:

- hosts never spread out, although `acquire_for`'s own doc says *"Different profiles or hosts spill into other slots or evict the coldest"*;
- concurrent jobs (parallel tier-2 escalations, `maybe_pre_solve` racing a fetch, the search ghost lane) serialize behind slot 0's lock while slots 1..N stay empty — the "other slots stay free for concurrent acquires" property never materializes;
- the host-affinity hint degenerates to an always-overwritten label on one slot.

Net effect: the pool never pools; behavior is the legacy single slot with extra bookkeeping.

There is a second, compounding problem: the slot's claim reached the meta snapshot only **after** `Ghost::launch` completed (seconds later). Concurrent acquires all read the same pre-launch snapshot, picked the same slot, and stacked behind one launch even where spilling was intended.

## The fix

1. **`pick_slot` ranking reworked** (still pure/testable): exact persona+host match → same-persona with no *competing* affinity (a hostless job or hostless slot rides along with anything) → compatible held-empty same-persona → free slot (**the spill**) → coldest same-persona warm slot (a thaw beats a relaunch and beats killing a stranger's browser) → coldest eviction.
2. **`claim_slot`**: pick and stamp `key/host/used` in one step under the meta lock, before the launch. A same-host peer still joins the in-flight claim and warm-serves behind it; a different host takes the next free slot.

All six previous selection tests pass unchanged; three new tests pin the spill, the exhausted-pool preference (reuse own coldest before evicting a stranger), and claim-time spreading. Also removes a copy-pasted duplicate doc block on `pool_slots`.

## RED → GREEN

The three new tests against the unmodified code:

```
different_hosts_spill_into_free_slots        FAILED  (picked 0, "a new host must open a free slot")
exhausted_pool_reuses_coldest_own_slot...    FAILED  (picked 0, first same-persona hit)
concurrent_claims_spread_hosts...            FAILED  (second host stacked on slot 0)
```

After: `cargo nextest run` **1085/1085 passed** (1 skipped), clippy `--all-targets` clean, fmt clean.

## Reviewed adversarially; residuals to be aware of

- **Claim vs reaper race** (µs window before the slot lock is taken vs the 5s reap tick): the reaper's browserless-slot reset can wipe a fresh claim; one acquire then degrades to the old stacking, self-heals. Not fixable without giving the reaper claim awareness; judged not worth the state.
- **Two different personas evicting concurrently** can still pick the same coldest slot and serialize — same shape as before this patch.
- **Deliberate behavior change to own**: the accidental funnel kept one Chrome alive; real spilling holds up to `DONSETCH_GHOST_POOL_SLOTS` (default 3) frozen Chromes within `REAP_AFTER`. That is the pool's stated design, and `DONSETCH_NO_GHOST_POOL` still reduces to exactly the legacy single-slot behavior (verified: the single incompatible-host slot is caught by the same-persona reuse rule — warm reuse, no relaunch).
- Verified against the call sites: all five daemon acquires share `host_default()`, and the scorecard's hostless `acquire()` rides along under the compatibility rule without killing a daemon browser (`guard.key`, the kill authority, is only written under the slot lock; meta claims never trigger kills).

Drive-by note, not in this PR: `xvfb.rs::xdpyinfo_probe`'s doc says *"False when xdpyinfo is missing"* but the code deliberately returns `true` (trust the connect check) — the doc sentence looks inverted.

### DIFF
--- a/src/ghost/manager.rs
+++ b/src/ghost/manager.rs
@@ -34,11 +34,6 @@ use crate::profile::BrowserProfile;
 /// default (a zero-slot pool would disable the warm path entirely,
 /// which the kill switch owns); `DONSETCH_NO_GHOST_POOL` forces the
 /// legacy single-slot path.
-/// Pool size: default 3 warm slots. Env override
-/// `DONSETCH_GHOST_POOL_SLOTS` clamps to 1..=16; 0 falls back to the
-/// default (a zero-slot pool would disable the warm path entirely,
-/// which the kill switch owns); `DONSETCH_NO_GHOST_POOL` forces the
-/// legacy single-slot path.
 fn pool_slots(default: usize) -> usize {
     pool_size(
         std::env::var_os("DONSETCH_NO_GHOST_POOL").is_some(),
@@ -69,10 +64,6 @@ struct Snap {
     key: Option<u64>,
     host: Option<String>,
     used: Instant,
-    /// Job in flight holds this slot's lock. The selector skips
-    /// busy slots so a concurrent same-persona hit spawns on a
-    /// free slot instead of serializing behind a warm one.
-    busy: bool,
 }
 
 struct Slot {
@@ -129,10 +120,6 @@ impl Drop for GhostGuard {
             && let Some(snap) = snaps.get_mut(self.idx)
         {
             snap.used = Instant::now();
-            snap.busy = false;
-            snap.live = self.guard.ghost.is_some();
-            snap.key = self.guard.key;
-            snap.host = self.guard.host.clone();
         }
         // On Windows and macOS, a frozen browser window stays visible
         // (Windows: taskbar, macOS: desktop). On Linux with Xvfb the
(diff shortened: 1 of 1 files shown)


dondai44423:
Reviewed in full, well caught: meta-level claims are the correct generalization of the e881922 busy-skip idea. Rebased onto current master so the branch carries both your spill/claim design and the recent selector changes; conflicts resolved by adopting your pick order verbatim and dropping the interim busy-flag mechanic in favor of claim_slot (claims are visible to pickers before the launch; the reaper still defers via try_lock). Also re-added the reap-countdown stamp at freeze that the rebase displaced. Your commit is preserved as its own authorship. Once CI is green this lands via rebase merge. Thanks for the test that pins the funnel.

PR #169, opened 2026-09-10
fix(transport): h3 exit rides finish(), honors alt-svc ma=, caps the body, protects routes.json

mnaza:
Four hardening fixes for the fresh h3 lane (3cf7a49 / 27456b2), one commit, all four verified against FETCH_HEAD.

## 1. The h3 exit bypassed `finish()` — hardcoded `Verdict::ContentOk`, no decompression

The h3 success arm hand-rolled a `FetchOutcome` with a literal `Verdict::ContentOk`. `finish()`'s own comment calls itself "one site of truth" for decompress + `walls::detect` (Q4) — the h3 lane added an N+1th exit that skipped both:

- **Decompression**: `finish()` is the *only* decompress site, so every gzip/br h3 body reached extraction — and the outer loop's marker scan — still compressed. (The h3 request filters `accept-encoding` out, but a server that compresses regardless went unhandled.)
- **Verdict**: the outer fetch loop re-scores `walls::detect` in its `_` arm, so the full `fetch()` flow healed the hardcode — but one-hop callers (`fetch_once_via`: search `engine_task`, enrich/prewarm) took `ContentOk` at face value. Concretely: enrich would park an h3-served Cloudflare challenge page into the prewarm store as real content, and the challenge re-fetch path never fired for those callers.

Fix: the exit now returns `finish(url, "h3", …)` like every other transport. A `finish()` error (undecodable payload) is treated as a transport failure: drop the route vouch, fall through to h1/h2 within the same call — no loop is possible (the h3 gate requires a route-memory hit, and the vouch is gone).

## 2. `ma=` parsed carefully, then thrown away

All three alt-svc absorb sites destructured `(port, _)` and recorded a hardcoded `86400`: a server vouching for 60 seconds was cached for a day — against routes.rs's own documented contract ("entries expire on `ma=`"), its `MA_DEFAULT = 3600` Chrome-default, and its parser tests that assert the extracted lifetime. New `routes::absorb_alt_svc(origin, value, egress)` records the server's parsed lifetime; the three call sites collapse onto it. Test pins `ma=0` → expired vouch never routes, `ma=600` → routes.

## 3. Unbounded h3 body

`body.extend_from_slice` in the `Event::Data` loop had no cap — a hostile origin could stream an unbounded body into RAM. This is the same remote-alloc class the h1 chunked reader was hardened against (#144), and h1 documents the rule: "bombs must fail before they allocate." `MAX_BODY` (64 MiB) moves to `transport/mod.rs`, shared by h1 and h3 so the two can never drift; `accept_body_chunk()` refuses the crossing chunk before extending. Boundary semantics match h1 (`> MAX_BODY` rejects, exactly `MAX_BODY` allowed — test pins both, plus "a refused chunk must not allocate"). Downstream, `decompress.rs`'s own `MAX_DECOMPRESSED` cap now applies to h3 bodies too via fix 1.

## 4. `routes.json` world-readable

The file carries serialized QUIC TLS sessions (0-RTT resumption material) and the module doc promises "same trust shape as the cookie vault" — but `save_file` wrote it via plain `fs::write` (0644 at umask default). It now writes through `config::write_private` (0600 from the first byte, the #139 primitive) to the tmp file, and the existing atomic rename preserves the mode.

## Verification

- routes perms test was **RED against the old `save_file`** (0644 observed), GREEN after.
- `absorb_alt_svc` + `accept_body_chunk` compile-RED → behavior pinned by unit tests.
- `finish()` contract pinned for the `h3` alpn: cf-mitigated 403 → `Challenge`, gzip body decompressed.
- Full `cargo nextest run`: **1094/1094 passed** (1 skipped); `clippy --all-targets` and `fmt --check` clean.
- Honest limit: a live h3 end-to-end rig is not feasible offline (real cert verification is non-optional in `build_quic_ctx_builder`), so the lane wiring is verified by inspection + the `finish()` contract test; adversarially reviewed (redirect/304 flow confirmed unchanged — the outer loop matches on `out.status` before verdict; h1/h2 header names confirmed lowercase at both absorb sites, h1.rs and hpack.rs).

Residual flagged, not fixed here: `h3_fetch_direct` builds its own `chrome_150(Platform::host())` profile instead of riding the Fetcher's configured profile — TLS-shape coherence question for a follow-up; and the reader loop is try_recv + 5–20 ms sleep polling rather than readiness-driven, a latency (not correctness) matter.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -60,8 +60,10 @@ skip-tree = []
 
 [sources]
 # Only crates.io + vendored paths (PDFium static libs are fetched
-# by build.rs, not a registry dep).
+# by build.rs, not a registry dep) + the donsetch quiche fork the
+# h3 lane pins by tag (3cf7a49; boring lifted to ^5 there so one
+# BoringSSL build serves h1/h2 and QUIC).
 unknown-registry = "deny"
 unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
-allow-git = []
+allow-git = ["https://github.com/dondai44423/quiche"]
--- a/src/fetch/client.rs
+++ b/src/fetch/client.rs
@@ -627,28 +627,31 @@ impl Fetcher {
             .await
             {
                 Ok((h3out, _stats)) => {
-                    if let Some(alt) = h3out.altsvc.as_ref()
-                        && let Some((port, _)) = crate::transport::routes::parse_h3_candidate(alt)
-                    {
-                        crate::transport::routes::record_h3(&origin, port, 86400, "direct");
+                    if let Some(alt) = h3out.altsvc.as_ref() {
+                        crate::transport::routes::absorb_alt_svc(&origin, alt, "direct");
                     }
                     if h3out.status == 0 || h3out.status < 200 {
                         crate::transport::routes::drop_h3(&origin);
                         // fall through to h1/h2
                     } else {
                         self.store_hop_cookies(use_jar, host, is_https, &h3out.headers);
-                        return Ok(FetchOutcome {
-                            url: url_str.to_string(),
-                            status: h3out.status,
-                            alpn: "h3".into(),
-                            headers: h3out.headers,
-                            body: h3out.body,
-                            redirects: 0,
-                            cache: CacheState::None,
-                            used_pool: false,
-                            verdict: Verdict::ContentOk,
-                            elapsed: std::time::Duration::from_millis(0),
-                        });
+                        // Same exit as every other transport: finish()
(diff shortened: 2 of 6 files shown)


mnaza:
Added a second commit: master's cargo-deny job has been red since 3cf7a49 — the quiche git pin landed while deny.toml still had `allow-git = []`, so every later master run fails `source-not-allowed` on quiche/octets. Allowed exactly that one repository (`unknown-git` stays `deny` for everything else). The 1 fail on this PR's first run was that inherited breakage, not the hardening diff.

PR #174, opened 2026-09-10
fix(cli): status used cfg! instead of #[cfg], breaking every default-feature build

Mart-Bogdan:
## Summary

`cfg!` is a runtime boolean, so the rerank-only arms stayed in the token stream and still had to resolve crate::memory::{kill_switch, rows, cap} — which src/memory/mod.rs gates behind #[cfg(feature = "rerank")]. Three E0425s, so `cargo build` with default features has failed since the web-memory commit.

Every red CI lane is a `features: ''` lane; the green ones all pass --features ocr,rerank[,http]. That also silenced the five fuzz targets (extract, charset, feed, sitemap, paginate), which build default-featured and had stopped running.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,20 @@ channel until the v4.0.0 release train.
 
 ### Fixed
 
+- Default-feature builds compile again: the web-memory receipt in
+  `donsetch status` tested the rerank feature with `cfg!` instead of
+  `#[cfg]`. `cfg!` is a runtime boolean, so the rerank-only arms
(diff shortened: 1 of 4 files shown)


Mart-Bogdan:
Hm, CI failed. I've tested locally iwth cargo clippy --all-targets --features ocr,rerank,http -- -Dwarnings but this 3 warnings exist only if `rerank` is off

PR #176, opened 2026-09-10
fix(fetch,memory): two daemon-killing paths in the v4 fetch and memory stacks

Mart-Bogdan:
## Summary

Both are unreleased v4 code, and both take down the whole MCP daemon rather than failing the request that hit them.

shadow.rs: scan_candidates searched a to_lowercase() copy of the document for tag offsets, then sliced the ORIGINAL string at them. to_lowercase is not byte-length preserving — İ 2->3, K 3->1, Ω 3->2 — so one such character shifted every later offset. Where the shift landed on ASCII the tag was mis-parsed and its assets silently dropped; where it did not, the slice started mid-codepoint and panicked, which the release profile's panic = "abort" turns into a daemon kill driven by page content. to_ascii_lowercase rewrites only 0x41-0x5A and leaves every byte >= 0x80 alone, so length and char boundaries are preserved and the offsets are valid by construction. It is also what HTML specifies: tag and attribute names match ASCII-case-insensitively. attr() carried the same pattern.

memory/model.rs: ensure_file returned ensure_file(..) with identical arguments when the file on disk failed its SHA/size pin, and never removed it, so every frame re-read the same bytes — a stack overflow, or an infinite loop if the recursion was optimized into a tail call. It repeated on every later call, so one corrupt model wedged every fetch that reached web-memory ingest. Remove the file and fall through to the existing download, which still installs by temp-write + verify
+ rename. A concurrent remover carries a receipt rather than an error; any other removal failure is immediate, since a sharing violation would fail the rename anyway after a 23MB download.

Regression tests for both. Against the old code the shadow input drops assets on Ω and panics on K, and the model test overflows the stack instead of reporting.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

## Breaking changes

None.

One deliberate narrowing worth naming: the scanner now folds tag and attribute
names with `to_ascii_lowercase` instead of `to_lowercase`, so a tag name written
with a non-ASCII character whose Unicode lowercase lands in ASCII — in practice
only `K` U+212A KELVIN SIGN — is no longer recognized. Browsers do not recognize
it either: the HTML standard matches tag and attribute names ASCII-case-
insensitively, so this drops an over-match rather than a capability.

Operationally visible: a locally cached embedding model that fails its SHA/size
pin is now deleted and redownloaded (~23 MB) instead of being retried in place.
Previously that path never terminated, so nothing depended on the old behavior.

## Test plan

Two regression tests, both discriminators against the old code:

- `fetch::shadow::tests::non_ascii_before_a_tag_does_not_shift_offsets` — a page
  carrying `İ` / `K` / `Ω` before a `<link>` and a `<script>`. Against the old
  scanner `Ω` shifts the offsets onto ASCII and both assets vanish, and `K`
  shifts onto a continuation byte and the slice panics.
- `memory::model::tests::corrupt_existing_file_is_removed_and_does_not_recurse` —
  seeds a file that fails the pin, then calls `ensure_file` against
  `http://127.0.0.1:1/` for an immediate connection refusal. Against the old code
  the recursion re-reads the same bytes every frame and the process dies on a
  stack overflow instead of the assertion reporting.

Run: `cargo nextest run --locked --features ocr,rerank,http` (filtered to the two
new tests plus the pre-existing `extracts_browser_subresources_in_order`, to catch
a regression in normal parsing) — 3 passed, 0 failed, on Windows x86_64 MSVC.

Full clippy and test suite run locally on this branch, clean. Not verified by
execution: the per-character behavior of the *old* code is traced from the byte
offsets rather than run, and neither failure mode was reproduced against a live
page or a genuinely corrupt model file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,25 @@ channel until the v4.0.0 release train.
 
 ### Fixed
 
+- Subresource shadow-fetching no longer aborts the daemon on a page
+  containing `İ`, `K` or `Ω`. The scanner searched a `to_lowercase()`
+  copy of the document for tag offsets and then sliced the original
+  string at them, but `to_lowercase` is not byte-length preserving, so
+  one such character shifted every later offset: assets silently
+  dropped where the shift landed on ASCII, a mid-codepoint slice panic
+  where it did not : and the release profile's `panic = "abort"` turns
+  that into a daemon kill. Folding is `to_ascii_lowercase` now, which
+  is byte-length and char-boundary preserving, and is also what the
+  HTML standard specifies for tag and attribute names. `attr()` had
+  the same latent pattern and is fixed with it.
+- A corrupt local embedding model no longer wedges every subsequent
+  fetch. `ensure_file` recursed into itself with identical arguments
+  when the file on disk failed its SHA/size pin, never removing it, so
+  the same bad bytes were read forever : a stack overflow, or an
+  infinite loop if the recursion was optimized into a tail call. The
+  bad file is removed and the existing atomic download path takes
+  over; a concurrent remover is tolerated with a receipt, any other
+  removal failure is honest and immediate.
 - Default-feature builds compile again: the web-memory receipt in
   `donsetch status` tested the rerank feature with `cfg!` instead of
   `#[cfg]`. `cfg!` is a runtime boolean, so the rerank-only arms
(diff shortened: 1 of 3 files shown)


Mart-Bogdan:
For some reason CI didn't start :-(

Mart-Bogdan:
Reopened as #179

Issue #173, opened 2026-09-10
`donsetch login --logout DOMAIN` no longer removes the session cookie (v4 `tier1_cookies`)


Mart-Bogdan:
> [!NOTE]
> ## **Disclaimer**
> I haven't dug deep into it, but I notice that lot of work is now happening on v4, so I've decided to bring this findings of my agent to your attentions, while you are working on it.
> Generally so much have changed in a project for last week that I haven't wrapped my head around it yet. 
> I am reporting this as it looks plausible, and importnat.

## `donsetch login --logout DOMAIN` no longer removes the session cookie (v4 `tier1_cookies`)

On current `master`, logout clears the two stores it knows about, but v4 phase 1.4 added a third
copy of the same cookies that logout never touches — and the daemon imports that third copy back
into the jar on the next start. The session survives logout.

This is on `master` only; v3.6.7 and earlier are unaffected. Flagging it now because it would ship
in v4.0.0.

### Where the third copy comes from

`GhostState::sync_tier1_cookies` (`src/ghost/cache.rs:1033`) appends the whole tier-1 jar into a
flat, top-level `tier1_cookies` array in `ghost-state.json`. It is called after every completed
navigation:

```rust
// src/mcp/server/fetch_tool.rs:829
state.sync_tier1_cookies(&daemon.fetcher.jar_all_snapshot().await);
```

The jar it snapshots already contains the vault's session cookies, because the daemon imports them
at boot (`src/mcp/server/mod.rs:80-86`). So one authenticated fetch is enough to copy a login cookie
out of the vault and into `tier1_cookies`.

### Why logout misses it

`clear_session_cookies_for` only walks `profiles[*].session_cookies`:

```rust
// src/ghost/cache.rs:640-649
let mut state = GhostState::load();
let mut removed = false;
for p in state.profiles.values_mut() {
    let before = p.session_cookies.len();
    p.session_cookies.retain(|c| { /* domain match */ });
    removed |= p.session_cookies.len() != before;
}
```

`AuthRegistry::remove` (`src/auth.rs:138-146`) calls exactly that one function, so the top-level
`tier1_cookies` array is never considered.

### Why it comes back

```rust
// src/mcp/server/mod.rs:183-191  (refresh_vault)
let mut cookies = crate::ghost::cache::load_session_cookies();
if !crate::config::env_flag("DONSETCH_NO_COOKIE_VAULT") {
    cookies.extend(self.state.lock().await.tier1_cookies.clone());
}
[...]

PR #188, opened 2026-09-11
fix(cookies): logout also wipes the rendered-DOM echo of the session (#173 deeper pass)

mnaza:
## The gap

`c346a26` (#173) closed the tier-1 cookie echo that logout was missing. But the session leaves a **fourth** copy of its fruit that logout still doesn't touch: the **render cache**.

A page fetched behind a session is stored whole in `GhostState.renders` (keyed by URL, `RENDER_TTL = 5 min`) and served by the tier-2 fetch shortcut — `render_for(url)` on the fetch and search paths. So after:

```
donsetch login --logout example.com
```

a fetch of a page that was rendered while logged in returns the **authenticated DOM straight from cache**, no network hop, for up to five minutes. The logged-out dashboard is handed back as if the session were still live. It's not a live credential (it can't authenticate new requests, and it self-expires in ≤5 min), so this is a content echo, not a full auth bypass — but it directly contradicts the logout contract, which is exactly the class of leak #173 set out to close.

## The fix

`clear_session_cookies_for` now sweeps `renders` with the same rule it already applies to the two cookie buckets: drop every render whose host belongs to the logged-out domain (`search::rank::host_of` + `auth::cookie_belongs_to`), keep everything else.

```rust
let before_r = state.renders.len();
state.renders.retain(|url, _| {
    let host = crate::search::rank::host_of(url);
    !crate::auth::cookie_belongs_to(&key, &host)
});
removed |= state.renders.len() != before_r;
```

`host_of` parses via the `url` crate (host lowercased) and `cookie_belongs_to` is the same bidirectional apex/subdomain match the cookie sweep uses, so `--logout example.com` clears `app.example.com`'s render and a subdomain logout clears its apex-fed render, while unrelated domains are left alone.

## RED → GREEN

Coverage is folded into the existing `logout_wipes_the_tier1_echo` test **on purpose, not as a new test**: the `auth_login` suite shares one on-disk `ghost-state.json` through a static `DONSETCH_CACHE_DIR`, so a second concurrent state writer races the whole-file save under plain `cargo test` (CI's nextest is process-per-test and immune, but a new flaky test helps no one). The one test now records a matching render (`app.alpha.test`) and an unrelated one (`beta.test`), logs out `alpha.test`, and asserts:

```
the rendered DOM of a logged-out domain must not survive logout   # RED before this fix
logout must not evict renders of unrelated domains
```

Verification: RED confirmed on the unmodified sweep (the `app.alpha.test` render survived); GREEN after. Full `cargo nextest run` **1105/1105 passed** (1 skipped); `clippy --all-targets` and `fmt --check` clean; `auth_login` green 3× under nextest.

## Found in the same pass, reported separately (not in this PR)

- **`web_screenshot` CLI drift**: `spec.rs` declares the tool with `cli_cmd: "screenshot"` and `donsetch screenshot URL` examples (rendered in `--help`), but `main.rs` never dispatches `screenshot`, so the spec's own example returns `unknown command 'screenshot'`. Fix direction is a design call (wire the CLI vs make it MCP-only per the commit's "CLI-free surface"), so I'm flagging rather than guessing.
- **`memory::store::persist` shared tmp**: the per-PID tmp (`index.PID.tmp`, `O_TRUNC` + rename) is now written by overlapping `ingest_async` tasks (fired per-256-chunk and on completion within one crawl); two concurrent persists can truncate/interleave one inode and the next `load()` parse-fails to `unwrap_or_default()` (a silent full index wipe). Real by construction; I could not force a deterministic repro locally (fast tmpfs), so I'm reporting it with the trivial fix (unique tmp per call) rather than shipping an unverified corruption claim.

### DIFF
--- a/src/ghost/cache.rs
+++ b/src/ghost/cache.rs
@@ -657,6 +657,18 @@ pub fn clear_session_cookies_for(domain: &str) -> bool {
         });
         removed |= p.session_cookies.len() != before;
     }
+    // Issue #173 (deeper pass): the rendered-DOM cache is a fourth copy
+    // of the session's fruit. A page fetched behind the session is
+    // stored whole in `renders` (RENDER_TTL = 5 min) and served by the
+    // tier-2 fetch shortcut, so a logged-out domain's dashboard would
+    // still hand back its authenticated DOM for up to five minutes.
+    // Drop every render whose host belongs to the logged-out domain.
+    let before_r = state.renders.len();
+    state.renders.retain(|url, _| {
+        let host = crate::search::rank::host_of(url);
+        !crate::auth::cookie_belongs_to(&key, &host)
+    });
+    removed |= state.renders.len() != before_r;
     if removed {
         state.save();
     }
--- a/tests/auth_login.rs
+++ b/tests/auth_login.rs
@@ -403,8 +403,17 @@ async fn probe_domain_does_not_panic_inside_a_tokio_runtime() {
 }
 
 /// Tier-1 jar echo hygiene (v4 phase 1.4, issue #173): the logout
-/// sweep owns every persisted copy. Two records, one matches the
-/// domain, one stays for another host.
+/// sweep owns every persisted copy. Two cookie records and two
+/// rendered pages, one of each matching the domain, one of each for
+/// another host. The rendered-DOM cache is a fourth copy of the
+/// session's fruit (a page fetched behind the session is stored
+/// whole in `renders`, RENDER_TTL = 5 min, and served by the tier-2
+/// fetch shortcut); a logout that leaves it hands back the logged-in
+/// DOM with no network hop for up to five minutes.
+///
+/// One test carries both echoes on purpose: the suite shares one
+/// on-disk state via the static cache dir, so a second concurrent
+/// writer would race the whole-file save under `cargo test`.
 #[test]
 fn logout_wipes_the_tier1_echo() {
     isolate_state();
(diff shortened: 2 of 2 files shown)


Mart-Bogdan:
Hello, I'll try to take a look.

Also your attached Claude session link doesn't work:
image

Mart-Bogdan:
Thanks for fixing this.

Two notes: one on the patch itself, one confirming your second side-finding.

## 1. `host_of` fails open, but the sweep should fail closed

`search::rank::host_of` returns an empty string when `Url::parse` fails or the URL has no host. An empty host matches none of the three arms of `cookie_belongs_to`:

- `"" == domain` → false
- `"".ends_with(".{domain}")` → false
- `domain.ends_with(".")` → false

So the render is **kept**. Every other branch of `clear_session_cookies_for` drops an entry when it matches; this one keeps an entry when it *cannot tell*.

To be clear about severity: I could not find a real input that produces such a key today. `renders` is keyed by URLs that were already parsed successfully in order to be fetched, so this is a defensive point, not a reproducible bug. What bothers me is the contract: `host_of` is a ranking helper, where an empty host is just a harmless scoring miss. Now it is load-bearing inside a logout sweep.

Keeping only hosts we could actually read preserves the fail-closed posture:

```rust
state.renders.retain(|url, _| {
    let host = crate::search::rank::host_of(url);
    !host.is_empty() && !crate::auth::cookie_belongs_to(&key, &host)
});
```

Dropping an unreadable key on logout costs nothing: it is a five-minute cache and re-renders on the next fetch.

## 2. `memory::store::persist` shared tmp file: confirmed on master

Since you could not force a repro, here is what I verified by reading the code:

- `persist` (`store.rs:117`) writes to `index.<pid>.tmp`, i.e. one tmp path per **process**, not per call.
- `ingest_batch` releases the store lock (`drop(acc)`) *before* calling `persist`, so nothing serializes two concurrent persists.
- `load()` (`store.rs:99-105`) is `.ok().and_then(|s| from_str(..).ok()).unwrap_or_default()`, so a torn index parses to an empty one with no error.

That matches your description: a silent full wipe of the web memory.

Issue #178, opened 2026-09-10
`web_memory` ingest runs inline on every fetch, outside `deadline_ms`, and rewrites the whole index each time


Mart-Bogdan:
> [!NOTE]
> ## **Disclaimer**
> I haven't dug deep into it, but I notice that lot of work is now happening on v4, so I've decided to bring this findings of my agent to your attentions, while you are working on it.
> Generally so much have changed in a project for last week that I haven't wrapped my head around it yet. 
> I am reporting this as it looks plausible, and importnat.

## `web_memory` ingest runs inline on every fetch, outside `deadline_ms`, and rewrites the whole index each time

Ingest is unconditional on `master` — the only control is the `DONSETCH_NO_WEB_MEMORY` opt-out. It fires at three tool surfaces:

- `src/mcp/server/fetch_tool.rs:92` (single) and `:214-227` (per URL in a batch)
- `src/mcp/server/crawl_tool.rs:105-112` — every crawled page
- `src/mcp/server/search_tool.rs:187-196` — every result row

**It is outside the budget.** `fetch_tool.rs:84-93` awaits `run_with_budget`, then calls `memory_ingest_result` before returning. A fetch that honored `deadline_ms` can still spend unbounded time afterward, and MCP cancellation doesn't reach it either.

**Each ingest rewrites the entire index.** `store::ingest` embeds, then `persist` (`src/memory/store.rs:107-121`) clones the whole `Vec<Entry>`, serializes it, and writes + renames the file. At the 4000-row default that's roughly 23 MB written *per page*. One 10-result search does it ten times sequentially; a 200-page crawl is on the order of gigabytes of writes, inline in the crawl worker.

**First use downloads mid-request.** Nothing pre-warms the model, so the first fetch after upgrading reaches `model::ensure_loaded` → `ensure_file`, which parks the thread on a blocking `recv` for a ~23 MB download with a 90s-per-file timeout — again after the deadline has been honored.

Suggestions, in rough order of payoff: move ingest off the request path (spawn or a queue), make `persist` an incremental write rather than a full rewrite, and pre-warm or lazily defer the first model download rather than paying it inside a fetch.

dondai44423:
Fixed in 677b1ce: ingest_batch (one embed pass + one persist per tool call), crawl chunks at 256, and all call sites hand the batch to tokio::task::spawn_blocking after the response so memory never holds it past deadline_ms.

PR #191, opened 2026-09-11
fix(memory): give each index persist its own staging file (concurrent-write corruption)

mnaza:
## The race

`677b1ce` (#178) moved web-memory ingest onto the blocking pool via `ingest_async`, and a single crawl now fires it **twice concurrently** — once per 256-row chunk during the crawl (`crawl_tool.rs`) and again with the remainder on completion. Search and fetch tools add more `ingest_async` callers. So `persist()` runs from overlapping `spawn_blocking` tasks.

`persist()` staged the whole index to a tmp keyed on the **PID alone** and wrote it with `O_TRUNC`, then renamed:

```rust
let tmp = path.with_extension(format!("{}.tmp", std::process::id()));  // same path every call
std::fs::write(&tmp, &body)?;   // O_CREAT | O_TRUNC
std::fs::rename(&tmp, &path)?;
```

Two concurrent persists open the **same** tmp. The second's `O_TRUNC` truncates the file the first is still writing; the two writers interleave one inode; whichever `rename` runs first moves the torn bytes into `index.json`. `load()` then hits `serde_json::from_str(...).ok()` → `unwrap_or_default()` and returns an **empty index** — the user's whole accumulated recall cache silently wiped on the next daemon start. (The second `rename` also fails ENOENT, swallowed by `ingest_async`'s `eprintln!`.)

The per-PID tmp was correct while ingest ran inline in the request task (persists never overlapped); it did nothing once `ingest_async` made them concurrent.

## The fix

Each persist stages to its **own** file — PID plus an atomic counter — so overlapping writers never share an inode, and every `rename` publishes a complete file from some writer:

```rust
fn stage_path(path: &Path) -> PathBuf {
    static SEQ: AtomicU64 = AtomicU64::new(0);
    let n = SEQ.fetch_add(1, Ordering::Relaxed);
    path.with_extension(format!("{}.{}.tmp", std::process::id(), n))
}
```

`rename` is atomic, so `index.json` is always a whole, valid file. The PID stays in the name so a second process never collides either.

## Verification — and an honest limit

This is a data race, so a deterministic corruption repro needs a slow filesystem: on a fast tmpfs the ~MB `write_all` completes as a burst before the `rename` window opens, and I could **not** force the interleave locally (32 threads × 60 iterations × 4 MB stayed clean). Rather than ship an unverified "reproduced a wipe" claim, the fix rests on removing the shared mutable staging file — an unambiguous antipattern once the writers became concurrent — and the tests pin what is deterministically checkable:

- `stage_paths_are_unique_per_write` — the invariant the fix guarantees (two calls stage to distinct files in the same dir); this is what the old per-PID path violated.
- `concurrent_persist_leaves_a_valid_index` — a best-effort guard that 8 threads hammering `persist()` always leave a parseable, complete index.

Full `cargo nextest run --features rerank` **1133/1133** (1 skipped); `clippy --all-targets` and `fmt --check` clean.

Residual, not addressed here (self-healing, minor): with `persist` deliberately outside the store lock (for latency), a slower writer's older snapshot can still be the last `rename`, dropping a row on disk until it is re-ingested. That is a lost row, not a corrupt file — the corruption/total-wipe path is what this PR closes. Closing the row-loss too would mean re-snapshotting under a writer lock, a larger change I left out.

### DIFF
--- a/src/memory/store.rs
+++ b/src/memory/store.rs
@@ -13,6 +13,7 @@
 //! the previous file intact.
 
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Mutex, OnceLock};
 use std::time::{SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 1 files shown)


dondai44423:
The staging-file fix itself is correct (adopted into master as part of the wave); the concurrency test as written ran persist() against the developer's real ~/.cache/donsetch memory index 160 times per run (no env isolation in nextest.toml, and cache_dir() honors DONSETCH_CACHE_DIR from the process), which our house rules forbid as a destructive test against real data. Landed on master with the same fix plus an in-test DONSETCH_CACHE_DIR isolating index_path; the PR could not carry the amendment since the head branch lives on your fork.

PR #192, opened 2026-09-11
fix(spec): web_screenshot is MCP-only, stop advertising a broken CLI subcommand

mnaza:
## The bug

`spec.rs` declares `web_screenshot` with `cli_cmd: "screenshot"` and three CLI examples:

```
donsetch screenshot https://example.com
donsetch screenshot https://example.com --full-page
donsetch screenshot https://example.com --wait-ms 2500
```

and `print_top_help` lists it under **AGENT TOOLS** in `donsetch --help`. But `main` dispatches only `fetch | search | answer | crawl | memory`, so `screenshot` falls through to the unknown-command arm:

```
$ donsetch screenshot https://example.com
donsetch: unknown command 'screenshot'
```

exit 1 — from the tool's **own advertised example**.

And the CLI surface was never going to work: `web_screenshot` returns an MCP image content block, and the CLI renderer (`render_result`) only emits `content[].text`, so it would drop the PNG entirely. The tool is MCP-only in practice, matching the commit that added it (“the CLI-free surface keeps the schema by the spec”). The spec just kept advertising a CLI it never had.

## The fix

Make it honest with a new `ToolSpec.mcp_only` flag. An MCP-only tool:
- is **not** resolved by `by_cli_cmd` (so it can never half-dispatch),
- is **not** listed in `donsetch --help`,
- carries no CLI examples (screenshot's are cleared to `&[]`).

It stays a first-class **MCP** tool: `tools/list` and the initialize `instructions` blurb still include `web_screenshot` unchanged — so `tools_list.json` and the `token_invariants` size gate are untouched (verified). The five real CLI tools set `mcp_only: false` and keep their subcommands exactly as before.

`main.rs` needs no change: `screenshot` was never dispatched; now it is simply no longer advertised, so nothing points a user at a command that doesn't exist.

## Verification

New test `mcp_only_tools_have_no_cli_surface_but_stay_on_the_mcp_surface`: `by_cli_cmd("screenshot")` is `None`, its examples are empty, `web_screenshot` is still present in `tools/list` by name, and every non-`mcp_only` tool still resolves as a CLI subcommand. Full `cargo nextest run` **1106/1106 passed** (1 skipped); `clippy --all-targets` and `fmt --check` clean; the tools/list fixture test and instructions gate pass unchanged.

### DIFF
--- a/src/cli/tool.rs
+++ b/src/cli/tool.rs
@@ -492,6 +492,10 @@ pub fn print_top_help() {
     println!();
     println!("AGENT TOOLS:");
     for tool in spec::TOOLS {
+        // MCP-only tools have no CLI subcommand to advertise here.
+        if tool.mcp_only {
+            continue;
+        }
         println!("  {:8} {}", tool.cli_cmd, tool.summary);
     }
     println!();
--- a/src/spec.rs
+++ b/src/spec.rs
@@ -81,8 +81,13 @@ pub struct ParamSpec {
 pub struct ToolSpec {
     /// MCP tool name (`web_fetch`).
     pub name: &'static str,
-    /// CLI subcommand (`fetch`).
+    /// CLI subcommand (`fetch`). Ignored when `mcp_only` is set.
     pub cli_cmd: &'static str,
+    /// MCP-surface-only tool: announced in `tools/list` and the
+    /// initialize `instructions`, but with no CLI subcommand — not
+    /// dispatched from `main`, not resolved by `by_cli_cmd`, and not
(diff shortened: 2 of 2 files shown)


dondai44423:
Superseded: the CLI subcommand now actually dispatches (it was spec'd but never wired, two match arms in main.rs gained 'screenshot') and the CLI gained a real --out PATH flag that writes the PNG after the capture, so the spec's examples no longer advertise a broken command. No doc-only workaround needed.

PR #196, opened 2026-09-11
feat(config): one typed runtime config, layered (closes #193)

### LINKED ISSUE #193: Proposal: consolidate runtime configuration for v4 (one TOML file + one env-var prefix)
# Proposal: consolidate runtime configuration for v4 (one TOML file + one env-var prefix)

## Pitch

DonSeTch has grown ~70 runtime knobs, all read straight from the process environment with `std::env::var*` at the point of use. They were added one feature at a time, and it shows:

- **Five prefixes for one binary.** `DONSETCH_*`, `DONSEEK_*`, `DONGHOST_*`, `DONSHEET_*`, `DONSIFT_*` are the fossil record of the old sub-crate names. A user has no way to guess which prefix a knob lives under. Some knobs are even read under two prefixes (`DONSETCH_BROWSER_BACKEND` falls back to `DONGHOST_BROWSER_BACKEND`).
- **Six different notions of "true".** Depending on the variable, "on" means: any value set (even empty), `1` only, `1|true|on`, `1|true|on|yes`, `is_ok()` (any value), or "not equal to `off`". `DONSETCH_NO_ADAPTERS=0` disables adapters; `DONSETCH_NO_H3=0` does nothing. Nobody can reason about that without reading the source.
- **Three ways to say "no".** Negative-named kill switches (`DONSETCH_NO_*`), positive names that accept `off` (`DONSHEET_OCR=off`, `DONSETCH_URL_HANDLES=off`), and positive names that accept `0|false|off|no|""` (`DONSETCH_BYPASS`). Plus positive/negative pairs where both exist (`DONSETCH_H3` + `DONSETCH_NO_H3`, `DONSETCH_SHADOW_FETCH` + `DONSETCH_NO_SHADOW_FETCH`).
- **Units and ranges are inconsistent.** Some knobs carry the unit in the name (`_SECS`, `_MS`, `_MB`), some do not (`DONSETCH_WEB_MEMORY_CAP`, `DONSETCH_XVFB_DISPLAY`). Some clamp silently, some ignore invalid input silently, one (`DONSEEK_RERANK_THREADS`) warns on stderr.
- **Only ~20 are documented.** The README mentions about a quarter of the list below. The rest are discoverable only by grepping.
- **Multi-harness deployments are painful.** People run DonSeTch under several MCP hosts (Claude Code, Claude Desktop, OpenCode, pi, ...) and CI containers. Each host has its own way of passing env to a server process, and each config blob has to repeat the same dozen exports. There is no way to say "this deployment uses profile X".

v4 is already a breaking release. It is the one window where renaming every knob costs nothing extra. After v4 ships, every rename needs a deprecation shim.

## Current inventory

Read from the v3.6.7 tree at the current v4 master (`b4de17f`). "Trigger" is what the code actually tests, not what the docs say.

Legend for the trigger column:
- **set** = any value, including empty string, enables it (`var_os(..).is_some()` / `var(..).is_ok()`)
- **flag** = `crate::config::env_flag`: `1`, `true`, `on` (case-insensitive, trimmed); everything else is false
- **=1** = exactly the string `1`
- **!=off** = enabled unless the value is exactly `off`

### Transport / MCP server

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONSETCH_TRANSPORT` | `http` (exact) | Start the HTTP MCP transport instead of stdio (only if the `http` feature is compiled in; otherwise exits with an error) | `src/main.rs` |
| `DONSETCH_HTTP_HOST` | string | Bind host (default `127.0.0.1`) | `src/main.rs` |
| `DONSETCH_HTTP_PORT` | u16 | Bind port (default `8765`, unparsable falls back silently) | `src/main.rs` |
| `DONSETCH_HTTP_TOKEN` | non-empty string | Bearer token; presence enables auth | `src/mcp/http.rs` |
| `DONSETCH_HTTP_TIMEOUT_SECS` | u64 | Per-request timeout | `src/mcp/http.rs` |
| `DONSETCH_HTTP_CORS` | `1`, `true`, `on` (case-sensitive) | Enable permissive CORS | `src/mcp/http.rs` |
| `DONSETCH_MCP_TEXT_ONLY` | flag | Never emit image blocks; text only | `src/mcp/compat.rs` |
| `DONSETCH_NO_ANSWER_TOOL` | flag | Hide / refuse the `web_answer` tool | `src/mcp/tools.rs`, `src/mcp/server/answer_tool.rs` |
| `DONSETCH_URL_HANDLES` | `!=off` | Short URL handles in search results (default on) | `src/handles.rs` |

### Paths / state

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONSETCH_CACHE_DIR` | non-empty path | Root for all on-disk state (cache, routes, handles, models, screenshots) | `src/paths.rs` |
| `DONSEEK_NO_DISK_STATE` | set | Skip persisting auth registry and route memory (test hook) | `src/auth.rs`, `src/ghost/cache.rs` |
| `DONSETCH_NO_ROUTE_MEMORY` | flag | Disable route memory (also disables probes and cookie vault) | `src/ghost/cache.rs`, `src/ghost/probe.rs`, `src/cli/status.rs` |
| `DONSETCH_ROUTE_MEMORY_READONLY` | flag | Read route memory but never write it | `src/ghost/cache.rs`, `src/ghost/probe.rs` |
| `DONSETCH_NO_COOKIE_VAULT` | flag | Do not persist tier-1 cookies | `src/ghost/cache.rs`, `src/mcp/server/mod.rs` |
| `DONSETCH_NO_WEB_MEMORY` | set | Disable the web memory index | `src/memory/model.rs`, `src/memory/store.rs` |
| `DONSETCH_WEB_MEMORY_CAP` | usize >= 256 | Max entries in the web memory index (smaller values ignored) | `src/memory/store.rs` |

### Fetch pipeline

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONSETCH_ALLOW_PRIVATE_EGRESS` | flag | Allow fetching loopback / RFC 1918 targets (SSRF guard off) | `src/fetch/guards.rs` |
| `DONSETCH_NO_ENV_PROXY` | flag | Ignore `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` | `src/fetch/client.rs`, `src/cli/doctor.rs` |
| `DONSEEK_PROXIES` | comma-separated proxy URLs | Extra proxies merged over the proxy config file | `src/transport/proxy.rs` |
| `DONSETCH_H3` | flag | Enable the HTTP/3 lane (opt-in) | `src/fetch/client.rs` |
| `DONSETCH_NO_H3` | flag | Kill the HTTP/3 lane even if `DONSETCH_H3` is set | `src/fetch/client.rs` |
| `DONSETCH_NO_ALT_SVC` | flag | Ignore `alt-svc` headers | `src/transport/routes.rs` |
| `DONSETCH_SHADOW_FETCH` | flag | Force page-load realism (asset burst) for every host | `src/fetch/shadow.rs` |
| `DONSETCH_NO_SHADOW_FETCH` | flag | Never do the asset burst (wins over the positive flag) | `src/fetch/shadow.rs` |
| `DONSETCH_SHADOW_DEADLINE_MS` | u64 > 0 | Wall clock for the asset burst (default 3000) | `src/fetch/shadow.rs` |
| `DONSETCH_NO_ADAPTERS` | set | Disable site adapters (Reddit, wiki infobox, ...) | `src/adapters/mod.rs`, `src/cli/adapters.rs` |
| `DONSETCH_ADAPTER_DUMP` | directory path | Dump every adapter-inspected body to that dir (fixture capture) | `src/adapters/mod.rs` |
| `DONSETCH_NO_CRAWL_SHAPE` | flag | Disable crawl output shaping | `src/crawl/mod.rs` |
| `DONSETCH_NO_PREWARM` | flag | Disable search-result prewarm fetches | `src/search/enrich.rs` |
| `DONSETCH_PDF_MAX_MB` | usize | Max PDF size accepted for parsing | `src/pdf/mod.rs` |
| `DONSHEET_OCR` | `!=off` | OCR tier for scanned PDFs (default on when the feature is built) | `src/pdf/ocr.rs` |
| `DONSHEET_OCR_MAX_PAGES` | usize | Max OCR pages per document (default 25) | `src/pdf/ocr.rs` |

### Bot-wall bypass (Bright Data unlocker)

All read in `src/fetch/bypass.rs` unless noted.

| Variable | Trigger / value | Effect |
|---|---|---|
| `DONSETCH_BYPASS` | `0`, `false`, `off`, `no`, `""` disables | Master switch (default on) |
| `DONSETCH_BYPASS_MAX_DAILY` | u32, clamped 1..=10000 | Daily unlock budget (default 50) |
| `DONSETCH_BYPASS_TIMEOUT_SECS` | u64, clamped 5..=600 | Unlock request timeout (default 120) |
| `DONSETCH_BYPASS_RENDER` | `1`, `true`, `on`, `yes` | Ask the unlocker to JS-render |
| `DONSETCH_BYPASS_ENDPOINT` | non-empty URL | Override the unlocker endpoint |
| `DONSETCH_BYPASS_CACHE` | `0`, `false`, `off`, `no`, `""` disables | Cache unlocked bodies (default on) |
| `DONSETCH_BYPASS_CACHE_TTL_SECS` | u64, clamped 0..=31536000 | Cache TTL (default 21600) |
| `DONSETCH_BYPASS_CACHE_MAX_ENTRIES` | u32, clamped 1..=100000 | Cache size (default 200) |
| `DONSETCH_UNLOCKER_ZONE` | non-empty string | Default Bright Data zone when the key has no `::zone` suffix |
| `DONSETCH_BRIGHTDATA_ZONE` | non-empty string | Same idea for the BYOK search engine (`src/search/byok/brightdata.rs`) |

### Search

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONSETCH_GOOGLE_PROFILE` | profile id | Pin the Google WML user-agent profile | `src/search/engines/google_wml.rs` |
| `DONSEEK_FORCE_GHOST_LANE` | set | Always run the Google ghost (browser) lane | `src/search/mod.rs` |
| `DONSEEK_NO_GHOST_LANES` | set | Never run the ghost lane | `src/search/mod.rs` |
| `DONSEEK_NO_TOPUP` | set | Skip the post-enrichment rerank top-up (A/B hook) | `src/search/mod.rs` |
| `DONSEEK_RERANK_THREADS` | usize 1..=MAX | ONNX intra-op threads for the reranker; invalid values warn on stderr and fall back | `src/search/rerank.rs` |

### Browser (tier 2) and ghost pool

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONSETCH_BROWSER_BACKEND` | `auto`, `chromium`/`chrome`/`original`, `headless`/`original-headless`, `cloak`/`cloakbrowser` | Which browser backend to launch; falls back to `DONGHOST_BROWSER_BACKEND` | `src/ghost/cloak.rs` |
| `DONGHOST_BROWSER_BACKEND` | same | Legacy alias of the above | `src/ghost/cloak.rs` |
| `DONGHOST_CHROME` | executable path | Explicit Chromium binary | `src/ghost/mod.rs`, `src/ghost/cloak.rs`, `src/profile/mod.rs` |
| `DONGHOST_NO_SANDBOX` | `=1` | Pass `--no-sandbox` to Chromium | `src/ghost/mod.rs` |
| `DONSETCH_CLOAK_AUTO_DOWNLOAD` | `=1` | Allow downloading the signed CloakBrowser binary | `src/ghost/cloak.rs` |
| `CLOAKBROWSER_BINARY_PATH` | executable path | Explicit CloakBrowser binary | `src/ghost/cloak.rs` |
| `CLOAKBROWSER_VERSION` | `a.b.c.d[.e]` numeric | Which CloakBrowser release to fetch | `src/ghost/cloak.rs` |
| `CLOAKBROWSER_CACHE_DIR` | path | Where downloaded CloakBrowser lives (default `<cache>/cloakbrowser`) | `src/ghost/cloak.rs` |
| `DONSETCH_NO_GHOST_POOL` | set | Legacy single-slot browser path | `src/ghost/manager.rs` |
| `DONSETCH_GHOST_POOL_SLOTS` | usize, clamped 1..=16, `0` = default | Warm browser slots (default 3) | `src/ghost/manager.rs` |
| `DONSETCH_XVFB_DISPLAY` | u8 1..=254 | Xvfb display number on Linux (default 99) | `src/ghost/xvfb.rs` |
| `DONSETCH_NO_ROUTE_PROBES` | flag | Disable the background wall re-probe | `src/ghost/probe.rs`, `src/mcp/server/mod.rs` |
| `DONSETCH_PROBE_SCAN_SECS` | u64 > 0 | Probe scan interval (test hook) | `src/ghost/probe.rs` |
| `DONSETCH_PROBE_STALE_SECS` | u64 | Age before a walled host is re-probed (test hook) | `src/ghost/probe.rs` |
| `DONSETCH_LOGIN_FORCE` | set | Skip the "no display" check in `donsetch login` | `src/cli/login.rs` |

### Debug / diagnostics

| Variable | Trigger / value | Effect | Where |
|---|---|---|---|
| `DONGHOST_DEBUG` | set | stderr tracing for browser, h3, routes, memory model, MCP tools | ~30 sites across `src/ghost`, `src/transport`, `src/memory`, `src/mcp/server` |
| `DONSEEK_DEBUG` | set | stderr tracing for search engines, BYOK, plugins | `src/search/**` |
| `DONSHEET_DEBUG` | set | stderr tracing for PDF extraction | `src/pdf/**` |
| `DONSHEET_DEBUG_MATRIX` | set | Dump PDF transform matrices | `src/pdf/engine.rs`, `src/pdf/rotate.rs` |
| `DONSHEET_DEBUG_CHARS` | set | Dump PDF char peeks | `src/pdf/layout.rs` |
| `DONSHEET_DEBUG_WORDS` | set | Dump PDF word assembly | `src/pdf/layout.rs` |
| `DONSIFT_DEBUG` | set | stderr tracing for extraction scoring | `src/extract/score.rs` |
| `DONSETCH_DEBUG_ECHO` | flag | Echo profile scorecard | `src/profile/scorecard.rs` |

### Standard / third-party variables honored

Deployment choices (these get a TOML mirror, see proposal section 4):

| Variable | Read where | Effect |
|---|---|---|
| `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` (and lowercase) | `src/transport/proxy.rs` | Egress proxy, scheme-specific first, then `ALL_PROXY` |
| `NO_PROXY` / `no_proxy` | `src/transport/proxy.rs` | Comma-separated host suffixes and CIDRs that bypass the proxy |
| `SSL_CERT_FILE`, `SSL_CERT_DIR` | `src/transport/tls.rs`, `src/cli/doctor.rs` | Extra trust roots, loaded by DonSeTch itself (not by the TLS library behind its back) |
| `TZ` | `src/persona.rs` | Persona timezone (must contain `/`) |
| `LC_ALL`, `LANG` | `src/persona.rs` | Persona locale (`en_US.UTF-8` becomes `en-US`) |
| `NO_COLOR` | `src/cli/mod.rs` | Disable ANSI color in CLI output |
| `PLAYWRIGHT_BROWSERS_PATH` | `src/ghost/mod.rs` | Extra Chromium discovery root |

Platform discovery, kept as-is and never mirrored: `DISPLAY`, `WAYLAND_DISPLAY`, `HOME`, `XDG_CACHE_HOME`, `LOCALAPPDATA`, `ProgramFiles`, `ProgramFiles(x86)`, `PREFIX` (Termux), `PATH`. They describe the machine, not the deployment.

Not in scope either: build-time `option_env!` values (`DONSHEET_PDFIUM`, `DONSHEET_TARGET`, `DONSHEET_FEATURES`, `DONSHEET_GIT_HASH`), the `BLESS_MCP_FIXTURES` test hook, and `DONSETCH_PLUGIN` / `DONSETCH_PLUGIN_NAME`, which DonSeTch *sets* for plugin subprocesses rather than reads.

### Inconsistencies at a glance

- Prefixes: `DONSETCH_` (49), `DONSEEK_` (7), `DONGHOST_` (4), `DONSHEET_` (6), `DONSIFT_` (1), `CLOAKBROWSER_` (3).
- "True" semantics in use: `set` (any value, even empty), `flag` (`1|true|on`), exactly `=1`, `1|true|on` case-sensitive, `1|true|on|yes`.
- "False" semantics in use: `NO_*` name (17 variables), `!=off` (2), `0|false|off|no|""` (2).
- Redundant pairs: `DONSETCH_H3`/`DONSETCH_NO_H3`, `DONSETCH_SHADOW_FETCH`/`DONSETCH_NO_SHADOW_FETCH`, `DONSEEK_FORCE_GHOST_LANE`/`DONSEEK_NO_GHOST_LANES`, `DONSETCH_BROWSER_BACKEND`/`DONGHOST_BROWSER_BACKEND`, `DONSETCH_UNLOCKER_ZONE`/`DONSETCH_BRIGHTDATA_ZONE`.
- Silent behavior on bad input everywhere except `DONSEEK_RERANK_THREADS`.

### 1. One typed config struct, loaded once

Replace the scattered `std::env::var*` calls with a single `Config` struct built at startup (and reachable via a `OnceLock`). Every module reads from it instead of the environment. Tests get a `Config::for_test()` builder instead of the current `set_var` juggling that already forces the suite onto nextest.

Use the [`config`](https://crates.io/crates/config) crate with **only the `toml` feature** (its default enables json/yaml/ron/ini/json5 and a `convert-case` dep; none of that is needed):

```toml
[dependencies]
config = { version = "0.15", default-features = false, features = ["toml"] }
```

Layering, lowest to highest precedence:

1. Built-in defaults (`Config::default()`).
2. Standard third-party variables (`HTTPS_PROXY`, `SSL_CERT_FILE`, `TZ`, ...; see 4).
3. TOML file (see 2).
4. `DONSETCH_*` environment (see 3).
5. CLI flags where they already exist (`--http`, `--host`, `--port`).

Each layer is more specific to this launch than the one below it: the ambient environment is set for every program on the machine, the TOML is set for DonSeTch, the `DONSETCH_*` variable is set for this one host registration.

### 2. A TOML file

Default location: `<config_dir>/donsetch/donsetch.toml` (via `dirs::config_dir()`, i.e. `~/.config/donsetch/` on Linux, `%APPDATA%\donsetch\` on Windows, `~/Library/Application Support/donsetch/` on macOS). Missing file is fine.

One env var, `DONSETCH_CONFIG`, overrides the path. This is the piece that makes multi-harness deployments sane: each MCP host registration passes a single `DONSETCH_CONFIG=/path/to/claude-code.toml` (or `opencode.toml`, `ci.toml`) instead of a dozen exports, and the profiles live in one directory under version control. A missing file named explicitly by `DONSETCH_CONFIG` is an error, not a silent fallback.

Section layout mirrors the tables above (values shown are today's defaults):

```toml
[transport]
kind = "stdio"          # "stdio" | "http"
http_host = "127.0.0.1"
http_port = 8765
http_token = ""         # empty = no auth
http_timeout_secs = 300
http_cors = false

[mcp]
text_only = false
answer_tool = true
url_handles = true

[paths]
cache_dir = ""          # empty = platform default

[state]
route_memory = "on"     # "on" | "readonly" | "off"
cookie_vault = true
web_memory = true
web_memory_cap = 4000

[proxy]
from_environment = true # read HTTP_PROXY/HTTPS_PROXY/ALL_PROXY/NO_PROXY as the bottom layer
http = ""               # mirrors HTTP_PROXY
https = ""              # mirrors HTTPS_PROXY
all = ""                # mirrors ALL_PROXY
no_proxy = []           # mirrors NO_PROXY
pool = []               # replaces DONSEEK_PROXIES: ["socks5://host:port", ...]

[tls]
cert_file = ""          # mirrors SSL_CERT_FILE
cert_dir = ""           # mirrors SSL_CERT_DIR

[persona]
timezone = ""           # mirrors TZ; empty = UTC
locale = ""             # mirrors LC_ALL / LANG; empty = en-US

[cli]
color = true            # NO_COLOR set = false

[fetch]
allow_private_egress = false
h3 = false
alt_svc = true
shadow_fetch = "auto"   # "auto" | "always" | "never"
shadow_deadline_ms = 3000
adapters = true
adapter_dump_dir = ""
crawl_shape = true
prewarm = true
pdf_max_mb = 100
ocr = true
ocr_max_pages = 25

[bypass]
enabled = true
max_daily = 50
timeout_secs = 120
render = false
endpoint = ""
cache = true
cache_ttl_secs = 21600
cache_max_entries = 200
zone = ""               # replaces both *_ZONE vars

[search]
google_profile = ""
ghost_lane = "auto"     # "auto" | "always" | "never"
rerank_topup = true
rerank_threads = 0      # 0 = automatic

[browser]
backend = "auto"        # "auto" | "chromium" | "headless" | "cloakbrowser"
chromium_path = ""
no_sandbox = false
cloak_auto_download = false
cloak_path = ""
cloak_version = ""
cloak_cache_dir = ""
playwright_path = ""    # mirrors PLAYWRIGHT_BROWSERS_PATH
pool_slots = 3
xvfb_display = 99
route_probes = true
probe_scan_secs = 120
probe_stale_secs = 21600
login_force = false

[debug]
ghost = false
search = false
pdf = false
pdf_matrix = false
pdf_chars = false
pdf_words = false
extract = false
echo_scorecard = false
```

Three-state knobs (`route_memory`, `shadow_fetch`, `ghost_lane`) collapse the current positive/negative env-var pairs into one enum, so "both set" can no longer happen.

### 3. Environment variables, derived mechanically from the TOML keys

`config`'s `Environment` source maps `DONSETCH_<SECTION>__<KEY>` onto `section.key` (prefix `DONSETCH`, separator `__`). Every TOML key above therefore has exactly one env name, and the mapping is a rule rather than a lookup table:

| TOML | Env |
|---|---|
| `transport.kind = "http"` | `DONSETCH_TRANSPORT__KIND=http` |
| `fetch.allow_private_egress = true` | `DONSETCH_FETCH__ALLOW_PRIVATE_EGRESS=true` |
| `browser.backend = "cloakbrowser"` | `DONSETCH_BROWSER__BACKEND=cloakbrowser` |
| `debug.ghost = true` | `DONSETCH_DEBUG__GHOST=true` |
| `bypass.max_daily = 10` | `DONSETCH_BYPASS__MAX_DAILY=10` |

Rules that apply uniformly:

- **Booleans** accept `true|false|1|0|on|off|yes|no`, case-insensitive, trimmed. Anything else is a config error. "Variable merely present" never means true.
- **Negative names are gone.** `DONSETCH_NO_X` becomes `x = false`. The only exception worth keeping is none.
- **Units are in the key** (`_secs`, `_ms`, `_mb`) or the key is unitless because the value is a count.
- **Out-of-range and unparsable values fail loudly** at startup (stderr, non-zero exit for the CLI, an `initialize` error for the MCP server), not silently clamped or ignored. The `DONSEEK_RERANK_THREADS` behavior becomes the norm.
- **Lists** (`proxy.pool`, `proxy.no_proxy`) are comma-separated in env, arrays in TOML.
- **Secrets** (`transport.http_token`, proxy passwords) work in both but the README should point people to env or the existing keys store rather than a world-readable TOML.

`DONSETCH_CONFIG` itself is the one env var outside this scheme, because it has to be read before the config exists.

### 4. Standard variables become the bottom layer, with a TOML mirror

`HTTPS_PROXY`, `SSL_CERT_FILE` and friends keep working exactly as today for anyone who never touches the TOML. What changes is that each one now has a TOML key and a `DONSETCH_*` name above it, so a deployment can pin or override them without editing the ambient environment:

| Standard var | TOML key | `DONSETCH_*` form |
|---|---|---|
| `HTTP_PROXY` | `proxy.http` | `DONSETCH_PROXY__HTTP` |
| `HTTPS_PROXY` | `proxy.https` | `DONSETCH_PROXY__HTTPS` |
| `ALL_PROXY` | `proxy.all` | `DONSETCH_PROXY__ALL` |
| `NO_PROXY` | `proxy.no_proxy` | `DONSETCH_PROXY__NO_PROXY` |
| `SSL_CERT_FILE` | `tls.cert_file` | `DONSETCH_TLS__CERT_FILE` |
| `SSL_CERT_DIR` | `tls.cert_dir` | `DONSETCH_TLS__CERT_DIR` |
| `TZ` | `persona.timezone` | `DONSETCH_PERSONA__TIMEZONE` |
| `LC_ALL`, `LANG` | `persona.locale` | `DONSETCH_PERSONA__LOCALE` |
| `NO_COLOR` | `cli.color` (inverted) | `DONSETCH_CLI__COLOR` |
| `PLAYWRIGHT_BROWSERS_PATH` | `browser.playwright_path` | `DONSETCH_BROWSER__PLAYWRIGHT_PATH` |

Precedence, lowest to highest: standard variable, TOML, `DONSETCH_*`. A machine-wide `HTTPS_PROXY` applies unless the TOML says otherwise, and one host registration can still beat the TOML. This is the same "more specific wins" rule as the rest of the layering, so nothing new to learn.

How it fits the `config` crate: `Environment` only maps by prefix, so the bottom layer is a small custom `Source` (one required method, `collect()`, returning a `Map<String, Value>`). It reads the ten names above, writes them under the TOML keys, and stamps each value's origin with the variable it came from, so `config show` prints `proxy.https = http://... (from HTTPS_PROXY)`. Uppercase before lowercase, as today. About forty lines.

Two rules that make this behave:

- **Opting out is a switch, not a sentinel.** `proxy.from_environment = false` (the successor of `DONSETCH_NO_ENV_PROXY`) drops the standard-variable layer for the proxy keys entirely. There is deliberately no `proxy.https = ""` meaning "explicitly none": empty string as a magic value is one of the things this proposal removes. Implementation is a two-pass build: read the switch from TOML plus `DONSETCH_*`, then build the full config with or without the standard-env source. Five lines.
- **Resolved values are re-exported to child processes.** Tier-2 Chromium and BYOK plugin subprocesses read `HTTPS_PROXY`, `NO_PROXY` and `SSL_CERT_FILE` themselves. If the TOML sets a proxy that the ambient environment lacks, DonSeTch sets the standard variables on the child's environment from the resolved config, so the browser and plugins see the same egress as tier 1. Without this a TOML proxy would silently apply to the HTTP client only.

`tls.cert_file` and `tls.cert_dir` are safe to mirror because `src/transport/tls.rs` loads the environment bundle itself rather than relying on the TLS library reading `SSL_CERT_FILE` at the C level; the config value simply feeds that loader.

### 5. Migration

- v4 drops the old names outright. This is the breaking change, and v4 is the release that already breaks.
- `donsetch doctor` gains a check that lists any legacy `DONSETCH_*`/`DONSEEK_*`/`DONGHOST_*`/`DONSHEET_*`/`DONSIFT_*`/`CLOAKBROWSER_*` variable found in the environment and prints the new key for each. Cheap to build from the table above, and it turns a silent "my flag stopped working" into a one-line fix.
- `donsetch config show` prints the effective merged config with the source of each value (default / file / env / flag). This comes almost for free, on one condition: the defaults must be fed into `config` as the bottom source (`Config::try_from(&DonsetchConfig::default())`) rather than living only in a Rust `Default` impl. Then `Config` implements `Source`, so `cfg.collect()` returns the merged map of every key, and `Value::origin()` reports where each value came from (file path, env variable name, or the defaults layer). A recursive walk over that map is the whole command.
- The same walk generates the knob list. Because the env name is a rule (`DONSETCH_<SECTION>__<KEY>`), each key can print its TOML path, env name, default, and current value. `donsetch config show --markdown` (or a `just` recipe) emits the README table, so the docs cannot drift from the struct and no one maintains an inventory like the one above by hand again.
- The README env-var sections get replaced with one page: the annotated TOML above, plus the two mapping rules, plus the generated table.

### 6. Tests stop mutating the process environment

Today the suite has 132 `std::env::set_var` / `remove_var` sites across 19 files (`src/transport/proxy.rs` alone has 42). Every one of them is `unsafe` on edition 2024, and every one of them is process-global state: a test that sets `DONSETCH_CACHE_DIR` moves *every other test's* on-disk state for as long as it is set. That is why the suite only works under nextest (one process per test) and fails deterministically under plain `cargo test`. It is also why `src/crawl/tests.rs` still races even under nextest: it never sets the variable itself, so it writes to the real user cache dir and collides with whatever daemon is running on the machine.

With a typed config the tests construct the struct instead:

```rust
let cfg = DonsetchConfig {
    paths: Paths { cache_dir: tmp.path().into() },
    fetch: Fetch { allow_private_egress: true, ..Default::default() },
    proxy: Proxy { from_environment: false, ..Default::default() },
    ..Default::default()
};
```

and pass it to the code under test. No `unsafe`, no cleanup on the way out, no dependency on which runner is in use, and two tests with different settings can share a process.

The one design choice that decides whether this pays off: **code must take `&Config` (or a handle to it), not read a process-wide `OnceLock`.** A global `CONFIG.get()` sprinkled through the modules is the same global state as `std::env::var`, just spelled differently, and tests would be back to "set the global, run, restore". Constructors already exist for the big objects (`Daemon`, the fetch client, the search core, `GhostManager`), so threading a config reference through them is mostly mechanical. For the leaf helpers that today call `env_flag` from deep inside, a `cfg` parameter on the few entry points that reach them is enough.

Layering also makes the existing test hooks (`DONSETCH_PROBE_SCAN_SECS`, `DONSETCH_PROBE_STALE_SECS`, `DONSETCH_SHADOW_DEADLINE_MS`) ordinary fields with no special status: a test sets `browser.probe_scan_secs = 1` on its struct and moves on.

The `config` crate itself does none of this; it only builds the struct at startup. The test win comes from the struct being the single input, so it is the same work as sections 1 and 3, not an extra project.

### 7. Out of scope

- Platform discovery variables (`HOME`, `XDG_CACHE_HOME`, `LOCALAPPDATA`, `ProgramFiles*`, `PREFIX`, `PATH`, `DISPLAY`, `WAYLAND_DISPLAY`) stay as they are. They describe the machine, and `paths.cache_dir` already covers the one location people want to move.
- The keys store (`donsetch keys add ...`) and proxy config file keep their current formats. This proposal is about runtime knobs, not credentials.
- Hot reload. Read once at startup, as today.

## Why the `config` crate rather than hand-rolling

It gives layered sources (defaults, file, env), the `__` separator env mapping, typed deserialization through serde, and a per-value provenance API, in about 150 lines of glue. Hand-rolling the same thing is how the current 70-ish variable sprawl happened. With `default-features = false, features = ["toml"]` the added surface is `config` itself, its two required deps (`winnow`, `pathdiff`), and `toml` 0.9 with its four helper crates (`toml_parser`, `toml_writer`, `toml_datetime`, `serde_spanned`). `serde` is already in the tree; no yaml/json/ron parsers come along.

**EDIT:** I have edited proposal with better handling of Standard variables

dondai44423:
Closes #193.

One typed config in `src/config.rs`: 13 sections, four layers, later wins.

- compiled defaults (a bare `donsetch mcp` stays the law)
- legacy env names, exact historical trigger semantics preserved in a custom source (`legacy_layer()`), deprecated, cut at the v4 release
- `donsetch.toml` via `DONSETCH_CONFIG` or the config dir; `DONSETCH_NO_CONFIG_FILE=1` skips it; setting both is an error; unknown TOML keys and bad values are hard errors naming the file
- `DONSETCH_<SECTION>__<KEY>` env names, strict (bad bool/int fatal, unknown var warns)

Surface: `donsetch config show` with per-knob origins, `--markdown` reference table, `--legacy` old-name map, doctor warns about deprecated vars active in the shell with key mappings. ~90 scattered env reads converted to the struct; every converted read keeps its behavior.

Two bugs the fixture gate caught during the pass, now fixed with regression coverage: the memory kill_switch enable/disable inversion (web_memory dropped from tools/list), and unescaped pipes splitting the markdown reference table. Env-switching tests split to fit the frozen-once config under nextest.

Gate x2 green: 1154 passed + 1 skipped. Live battery against the real binary: all four layer origins visible, loud errors on bad file/value/range/conflict, doctor legacy listing, daemon boot off a TOML (`debug.ghost`), tier-1 live fetch.

Left for review per #193: the layering order, the DONSEEK_RERANK_THREADS graceful exception, the legacy cut at v4, and the plugin/fixture surfaces touched nowhere else.

### DIFF
--- a/src/cli/doctor.rs
+++ b/src/cli/doctor.rs
@@ -278,7 +278,11 @@ async fn check_network(fetcher: &Fetcher) -> CheckResult {
             // fetch: curl works via the env proxy, direct sockets
             // get reset. Surface the fix instead of a generic
             // "check your connection".
-            let proxy_env_set = [
+            // The advice must match what the daemon actually does:
+            // a proxy comes from the ambient env vars OR the [proxy]
+            // slots in donsetch.toml (from_env_for consults the
+            // config layer first; the pool lane adds proxy.pool).
+            let proxy_configured = [
                 "HTTPS_PROXY",
                 "https_proxy",
                 "HTTP_PROXY",
@@ -287,12 +291,19 @@ async fn check_network(fetcher: &Fetcher) -> CheckResult {
                 "all_proxy",
             ]
             .iter()
-            .any(|v| std::env::var_os(v).is_some());
-            let hint = if proxy_env_set {
-                "The environment exports a proxy and the direct fetch still failed: this looks like a TLS-intercepting egress network. donsetch honors the env proxy automatically (see 'Fetch egress'); if it still fails, export SSL_CERT_FILE=<the network's CA bundle> so the re-signed certificates verify, then re-run doctor."
+            .any(|v| std::env::var_os(v).is_some())
+                || {
+                    let p = &crate::config::cfg().proxy;
+                    !p.https.trim().is_empty()
+                        || !p.http.trim().is_empty()
+                        || !p.all.trim().is_empty()
+                        || !p.pool.is_empty()
+                };
+            let hint = if proxy_configured {
+                "A proxy is configured (env vars or [proxy] in donsetch.toml) and the direct fetch still failed: this looks like a TLS-intercepting egress network. donsetch honors it automatically (see 'Fetch egress'); if it still fails, export SSL_CERT_FILE=<the network's CA bundle> or set [tls] cert_file in donsetch.toml so the re-signed certificates verify, then re-run doctor."
                     .to_string()
             } else {
-                "Check your network connection and DNS. Behind an egress proxy? Export HTTPS_PROXY/HTTP_PROXY (donsetch honors them; NO_PROXY accepted).".into()
+                "Check your network connection and DNS. Behind an egress proxy? Set [proxy] https/http in donsetch.toml or export HTTPS_PROXY/HTTP_PROXY (NO_PROXY accepted).".into()
             };
             CheckResult::Fail(e.to_string(), hint)
         }
@@ -303,21 +314,25 @@ async fn check_network(fetcher: &Fetcher) -> CheckResult {
 /// and the two certificate stores the connector builds from.
 /// Local-only: no network, stays in fast mode.
 fn check_fetch_egress() -> CheckResult {
-    let kill = !crate::config::cfg().proxy.from_environment;
-    let resolved = if kill {
-        None
-    } else {
-        crate::transport::proxy::from_env_for("https://example.com")
-    };
+    let proxy = &crate::config::cfg().proxy;
+    let slot_set = !proxy.https.trim().is_empty()
+        || !proxy.http.trim().is_empty()
+        || !proxy.all.trim().is_empty();
+    // from_env_for already consults the [proxy] slots first and
+    // gates only the ambient env read on from_environment: call it
+    // unconditionally so the doctor's view matches the daemon's
+    // (a TOML proxy with from_environment = false used to be
+    // reported as direct egress).
+    let resolved = crate::transport::proxy::from_env_for("https://example.com");
     let (sys_roots, env_roots) = crate::transport::tls::trust_store_report();
     let cert_bundle = std::env::var_os("SSL_CERT_FILE").map(|p| p.to_string_lossy().into_owned());
 
     let mut bits = Vec::new();
     if let Some(p) = resolved {
-        bits.push(format!("egress via {} proxy {}:{} (env; SOCKS5 keeps TLS end-to-end, HTTP CONNECT gets the interception-safe handshake)", if p.is_http_connect() { "http" } else { "socks5" }, p.host, p.port));
+        bits.push(format!("egress via {} proxy {}:{} ({}; SOCKS5 keeps TLS end-to-end, HTTP CONNECT gets the interception-safe handshake)", if p.is_http_connect() { "http" } else { "socks5" }, p.host, p.port, if slot_set { "config" } else { "env" }));
     } else {
-        bits.push(if kill {
-            "direct egress (proxy.from_environment = false disables the env-proxy convention)"
+        bits.push(if !proxy.from_environment {
+            "direct egress (proxy.from_environment = false disables the env-proxy convention; [proxy] slots in donsetch.toml stay live)"
                 .into()
         } else {
             "direct egress (no proxy env vars; export HTTPS_PROXY/HTTP_PROXY to route fetches)"
--- a/src/cli/doctor.rs
+++ b/src/cli/doctor.rs
@@ -183,6 +183,26 @@ pub async fn run() {
     // 16. MCP client registration (detect + print blocks).
     print_mcp_section();
 
+    // 17. Legacy env vars (the pre-v4 names): still honored, but
+    // each one active in this shell gets ONE warning naming its
+    // config key. Cut at the v4 release.
+    let legacy: Vec<&'static str> = crate::config::legacy_vars_in_env();
+    if legacy.is_empty() {
+        report!("Legacy env vars", CheckResult::Pass("none set".to_string()));
+    } else {
+        let names: Vec<String> = legacy
+            .iter()
+            .map(|name| {
+                let (section, key) = crate::config::legacy_target_of(name);
+                format!("{name} -> {section}.{key}")
+            })
+            .collect();
+        report!(
+            "Legacy env vars",
+            CheckResult::Warn(format!("{} (deprecated, cut at v4)", names.join(", ")))
+        );
+    }
+
     // ── Self-healing pass (--fix) ───────────────────────────
     if fix {
         println!();
@@ -283,7 +303,7 @@ async fn check_network(fetcher: &Fetcher) -> CheckResult {
 /// and the two certificate stores the connector builds from.
 /// Local-only: no network, stays in fast mode.
 fn check_fetch_egress() -> CheckResult {
-    let kill = crate::config::env_flag("DONSETCH_NO_ENV_PROXY");
+    let kill = !crate::config::cfg().proxy.from_environment;
     let resolved = if kill {
         None
     } else {
@@ -297,7 +317,8 @@ fn check_fetch_egress() -> CheckResult {
         bits.push(format!("egress via {} proxy {}:{} (env; SOCKS5 keeps TLS end-to-end, HTTP CONNECT gets the interception-safe handshake)", if p.is_http_connect() { "http" } else { "socks5" }, p.host, p.port));
     } else {
         bits.push(if kill {
-            "direct egress (DONSETCH_NO_ENV_PROXY=1 disables the env-proxy convention)".into()
+            "direct egress (proxy.from_environment = false disables the env-proxy convention)"
+                .into()
         } else {
             "direct egress (no proxy env vars; export HTTPS_PROXY/HTTP_PROXY to route fetches)"
                 .into()
@@ -800,7 +821,7 @@ fn check_ocr_models() -> CheckResult {
     #[cfg(feature = "ocr")]
     {
         if !crate::pdf::ocr::enabled() {
-            return CheckResult::Warn("disabled (DONSHEET_OCR=off)".into());
+            return CheckResult::Warn("disabled (fetch.ocr = false)".into());
         }
 
         let dir = crate::pdf::ocr::ocr_cache_dir();
--- a/src/config.rs
+++ b/src/config.rs
@@ -1965,6 +1965,56 @@ pub fn origins(
     out
 }
 
+/// Knobs whose values can carry credentials. Proxy URLs embed
+/// user:password; `config show` masks the secret part so terminal
+/// scrollback and pasted bug reports never leak it.
+fn is_secret(section: &str, key: &str) -> bool {
+    section == "proxy" && matches!(key, "http" | "https" | "all" | "pool")
+}
+
+/// Mask the userinfo of a URL credential: scheme://user:pass@host
+/// becomes scheme://***@host. Values without userinfo are not
+/// secrets and show as-is; bare non-URL values (never expected in
+/// these fields) mask fully rather than risk a leak.
+fn mask_url_secret(url: &str) -> String {
+    if let Some(scheme_end) = url.find("://") {
+        let rest = &url[scheme_end + 3..];
+        if let Some(at) = rest.find('@') {
+            return format!("{}://***@{}", &url[..scheme_end], &rest[at + 1..]);
+        }
+        return url.to_string();
+    }
+    if url.is_empty() {
+        String::new()
+    } else {
+        "***".to_string()
+    }
+}
+
+/// `config show` display form for one knob: value_of output for
+/// everything, credentials masked for the secret proxy fields.
+fn masked_value(c: &DonsetchConfig, section: &str, key: &str, shown: String) -> String {
+    if !is_secret(section, key) {
+        return shown;
+    }
+    if key == "pool" {
+        let masked = c
+            .proxy
+            .pool
+            .iter()
+            .map(|p| format!("\"{}\"", mask_url_secret(p)))
+            .collect::<Vec<_>>()
+            .join(", ");
+        return format!("[{masked}]");
+    }
+    let raw = match key {
+        "http" => c.proxy.http.as_str(),
+        "https" => c.proxy.https.as_str(),
+        _ => c.proxy.all.as_str(),
+    };
+    format!("\"{}\"", mask_url_secret(raw))
+}
+
 /// Human-readable `config show` output: every knob, value, source.
 pub fn show_text(loaded: &Loaded) -> String {
     let mut out = String::new();
@@ -1991,7 +2041,10 @@ pub fn show_text(loaded: &Loaded) -> String {
             out.push_str(&format!("[{section}]\n"));
             last_section = section;
         }
-        out.push_str(&format!("  {key:<18} = {value:<20} ({origin})\n"));
+        out.push_str(&format!(
+            "  {key:<18} = {} ({origin})\n",
+            masked_value(&loaded.config, section, key, value)
+        ));
     }
     out
 }
@@ -2345,6 +2398,11 @@ mod tests {
     /// proxy.from_environment gates the AMBIENT proxy convention, not
     /// the config-file slots: an explicit TOML proxy must survive the
     /// kill switch (the old caller-side gate starved it).
+    ///
+    /// Isolation note: relies on nextest's process-per-test runner
+    /// (env mutation + the process-wide cfg() OnceLock: the first
+    /// cfg() call freezes the config layer, so keep this out of
+    /// shared-process runners).
     #[test]
     fn from_environment_off_keeps_toml_proxy_slots_alive() {
         let guard = clean_env();
@@ -2462,6 +2520,31 @@ mod tests {
             .find(|l| l.contains("`kind`"))
             .expect("transport.kind not in the table");
         assert!(kind_row.contains("\\|"), "kind row lost its escaped pipe");
+    }
+
+    #[test]
+    fn show_text_redacts_proxy_secrets() {
+        let mut c = DonsetchConfig::default();
+        c.proxy.https = "http://alice:s3cr3t@gw.example:2334".into();
+        c.proxy.all = "socks5://bob:hunter2@other.example:1080".into();
+        c.proxy.pool = vec!["http://carol:pw@a.example:8080".into()];
+        let loaded = Loaded {
+            config: c,
+            merged: config::Map::new(),
+            warnings: Vec::new(),
+            file: None,
+        };
+        let text = super::show_text(&loaded);
+        for leak in ["s3cr3t", "hunter2", "pw@a.example"] {
+            assert!(!text.contains(leak), "config show leaked {leak:?}");
+        }
+        for masked in [
+            "http://***@gw.example:2334",
+            "socks5://***@other.example:1080",
+            "http://***@a.example:8080",
+        ] {
+            assert!(text.contains(masked), "missing masked form {masked:?}");
+        }
     }
 
     #[test]
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,15 +1,2087 @@
-//! Small, shared parsers for process-level runtime configuration.
+//! Runtime configuration: one typed struct, layered sources, one home.
+//!
+//! Layers, lowest to highest precedence:
+//!   1. serde defaults (this file, matching the historical runtime defaults)
+//!   2. legacy env names (the pre-v4 scattered knobs, exact trigger semantics preserved)
+//!   3. `donsetch.toml` file layer (`DONSETCH_CONFIG` explicit, or the default location)
+//!   4. new env names: `DONSETCH_<SECTION>__<KEY>` (two underscores = one dot)
+//!
+//! Standard process vars (HTTP_PROXY, SSL_CERT_FILE, TZ, LANG, NO_COLOR,
+//! PLAYWRIGHT_BROWSERS_PATH, PATH, HOME, ...) stay ambient and are read at
+//! call time; the config value of the mirrored knob wins over the ambient
+//! var whenever it is set.
+//!
+//! Notes:
+//! - `DONSETCH_CACHE_DIR` is NOT a knob: `paths::cache_dir()` keeps reading
+//!   it per call so tests can isolate state; `paths.cache_dir` in the TOML
+//!   layer is the fallback when the env var is absent.
+//! - `DONSETCH_NO_CONFIG_FILE=1` skips the file layer entirely (hermetic
+//!   runs, CI).
+//! - Values from the new env layer or TOML fail the whole load when
+//!   unparsable or out of range (loud, with the key path named). Legacy
+//!   names keep their historical silent-fallback semantics, except that a
+//!   bad value now also emits a stderr warning.
+//! - `cargo test` users are broken by design: the global is frozen at first
+//!   use. Our runner is nextest (one process per test), which makes the env
+//!   layers behave exactly like before for every test.
 
-use std::ffi::OsStr;
+use config::Source as _;
+use std::sync::OnceLock;
 
-/// Read a security-sensitive opt-in flag from the environment.
-///
-/// Only explicit affirmative values enable the flag. Matching is
-/// ASCII-case-insensitive and ignores surrounding ASCII/Unicode
-/// whitespace; missing, non-Unicode, false, and unknown values all
-/// fail closed.
-pub(crate) fn env_flag(name: &str) -> bool {
+// ---------------------------------------------------------------------------
+// The struct
+// ---------------------------------------------------------------------------
+
+macro_rules! section {
+    ($name:ident { $($field:ident : $ty:ty = $default:expr),* $(,)? }) => {
+        #[derive(Debug, Clone, PartialEq, serde::Deserialize, serde::Serialize)]
+        #[serde(default, deny_unknown_fields)]
+        pub struct $name {
+            $(pub $field: $ty,)*
+        }
+        impl Default for $name {
+            fn default() -> Self {
+                Self { $($field: $default,)* }
+            }
+        }
+        impl $name {
+            /// Fold one layer into the accumulator: copy every field the
+            /// layer explicitly set (differing from the struct default).
+            pub(crate) fn fold_from(&mut self, other: &Self) {
+                let defaults = Self::default();
+                $( if other.$field != defaults.$field {
+                    self.$field = other.$field.clone();
+                } )*
+            }
+        }
+    };
+}
+
+section!(TransportSection {
+    kind: TransportKind = TransportKind::Stdio,
+    host: String = "127.0.0.1".into(),
+    port: u16 = 8765,
+    token: String = String::new(),
+    timeout_secs: u64 = 300,
+    cors: bool = false,
+});
+
+section!(McpSection {
+    text_only: bool = false,
+    answer_tool: bool = true,
+    url_handles: bool = true,
+});
+
+section!(PathsSection {
+    cache_dir: String = String::new(),
+});
+
+section!(StateSection {
+    no_disk_state: bool = false,
+    route_memory: RouteMemory = RouteMemory::ReadWrite,
+    cookie_vault: bool = true,
+    web_memory: bool = true,
+    web_memory_cap: usize = 4000,
+});
+
+section!(ProxySection {
+    from_environment: bool = true,
+    http: String = String::new(),
+    https: String = String::new(),
+    all: String = String::new(),
+    no_proxy: String = String::new(),
+    pool: Vec<String> = Vec::new(),
+});
+
+section!(TlsSection {
+    cert_file: String = String::new(),
+    cert_dir: String = String::new(),
+});
+
+section!(PersonaSection {
+    timezone: String = String::new(),
+    locale: String = String::new(),
+});
+
+section!(CliSection {
+    color: ColorMode = ColorMode::Auto,
+});
+
+section!(FetchSection {
+    allow_private_egress: bool = false,
+    h3: bool = false,
+    alt_svc: bool = true,
+    shadow_fetch: ShadowFetch = ShadowFetch::Auto,
+    shadow_deadline_ms: u64 = 3000,
+    adapters: bool = true,
+    adapter_dump_dir: String = String::new(),
+    crawl_shape: bool = true,
+    prewarm: bool = true,
+    pdf_max_mb: usize = 100,
+    ocr: bool = true,
+    ocr_max_pages: u32 = 25,
+});
+
+section!(BypassSection {
+    enabled: bool = true,
+    zone: String = String::new(),
+    max_daily: u32 = 50,
+    timeout_secs: u64 = 120,
+    render: bool = false,
+    endpoint: String = String::new(),
+    cache: bool = true,
+    cache_ttl_secs: u64 = 21_600,
+    cache_max_entries: u32 = 200,
+});
+
+section!(SearchSection {
+    google_profile: String = String::new(),
+    ghost_lane: GhostLane = GhostLane::Auto,
+    rerank_topup: bool = true,
+    rerank_threads: u32 = 0,
+    brightdata_zone: String = String::new(),
+});
+
+section!(BrowserSection {
+    backend: BrowserBackend = BrowserBackend::Auto,
+    chromium_path: String = String::new(),
+    no_sandbox: bool = false,
+    cloak_auto_download: bool = false,
+    cloak_path: String = String::new(),
+    cloak_version: String = String::new(),
+    cloak_cache_dir: String = String::new(),
+    playwright_path: String = String::new(),
+    pool_slots: u32 = 0,
+    xvfb_display: u16 = 0,
+    route_probes: bool = true,
+    probe_scan_secs: u64 = 120,
+    probe_stale_secs: u64 = 21_600,
+    login_force: bool = false,
+});
+
+section!(DebugSection {
+    ghost: bool = false,
+    search: bool = false,
+    pdf: bool = false,
+    pdf_matrix: bool = false,
+    pdf_chars: bool = false,
+    pdf_words: bool = false,
+    extract: bool = false,
+    echo_scorecard: bool = false,
+});
+
+#[derive(Debug, Clone, PartialEq, Default, serde::Deserialize, serde::Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct DonsetchConfig {
+    pub transport: TransportSection,
+    pub mcp: McpSection,
+    pub paths: PathsSection,
+    pub state: StateSection,
+    pub proxy: ProxySection,
+    pub tls: TlsSection,
+    pub persona: PersonaSection,
+    pub cli: CliSection,
+    pub fetch: FetchSection,
+    pub bypass: BypassSection,
+    pub search: SearchSection,
+    pub browser: BrowserSection,
+    pub debug: DebugSection,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum TransportKind {
+    #[default]
+    Stdio,
+    Http,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum RouteMemory {
+    Off,
+    #[default]
+    #[serde(alias = "readwrite")]
+    ReadWrite,
+    #[serde(alias = "readonly")]
+    ReadOnly,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum ShadowFetch {
+    #[default]
+    Auto,
+    Always,
+    Never,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum GhostLane {
+    #[default]
+    Auto,
+    Always,
+    Never,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum BrowserBackend {
+    #[default]
+    Auto,
+    #[serde(alias = "chrome", alias = "original")]
+    Chromium,
+    #[serde(alias = "headless", alias = "original-headless")]
+    Headless,
+    #[serde(alias = "cloakbrowser")]
+    Cloak,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Deserialize, serde::Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum ColorMode {
+    #[default]
+    Auto,
+    Never,
+    Always,
+}
+
+// ---------------------------------------------------------------------------
+// Loading
+// ---------------------------------------------------------------------------
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ConfigError {
+    Env(String),
+    File { path: String, message: String },
+    Validate(String),
+}
+
+impl std::fmt::Display for ConfigError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ConfigError::Env(m) => write!(f, "invalid env value: {m}"),
+            ConfigError::File { path, message } => {
+                write!(f, "invalid config file {path}: {message}")
+            }
+            ConfigError::Validate(m) => write!(f, "invalid config value: {m}"),
+        }
+    }
+}
+
+impl std::error::Error for ConfigError {}
+
+pub struct Loaded {
+    pub config: DonsetchConfig,
+    /// Non-fatal notes: ignored legacy values, unknown DONSETCH_* vars.
+    pub warnings: Vec<String>,
+    /// The file the TOML layer came from, if any.
+    pub file: Option<std::path::PathBuf>,
+}
+
+/// Load the layered config from defaults + env + TOML file and validate it.
+pub fn load() -> Result<Loaded, ConfigError> {
+    let mut warnings = Vec::new();
+
+    let mut file_layer: Option<config::Map<String, config::Value>> = None;
+    let mut file_path: Option<std::path::PathBuf> = None;
+    let skip_file = std::env::var_os("DONSETCH_NO_CONFIG_FILE").is_some();
+    if !skip_file {
+        match toml_path() {
+            Some(path) => {
+                let text = std::fs::read_to_string(&path).map_err(|e| ConfigError::File {
+                    path: path.display().to_string(),
+                    message: format!("unreadable ({e})"),
+                })?;
+                let source = config::File::from_str(&text, config::FileFormat::Toml);
+                file_layer = Some(source.collect().map_err(|e| ConfigError::File {
+                    path: path.display().to_string(),
+                    message: e.to_string(),
+                })?);
+                file_path = Some(path);
+            }
+            None => {
+                if let Some(explicit) = std::env::var_os("DONSETCH_CONFIG") {
+                    return Err(ConfigError::File {
+                        path: explicit.to_string_lossy().into_owned(),
+                        message: "not found".into(),
+                    });
+                }
+            }
+        }
+    } else if let Some(explicit) = std::env::var_os("DONSETCH_CONFIG") {
+        return Err(ConfigError::Env(format!(
+            "DONSETCH_NO_CONFIG_FILE=1 but DONSETCH_CONFIG={} is set; drop one of them",
+            explicit.to_string_lossy()
+        )));
+    }
+
+    let (legacy_map, legacy_warnings) = legacy_layer();
+    warnings.extend(legacy_warnings);
+    let (env_map, env_warnings, env_errors) = new_env_layer();
+    warnings.extend(env_warnings);
+    if let Some(e) = env_errors.first() {
+        return Err(ConfigError::Env(e.clone()));
+    }
+
+    let mut raw = DonsetchConfig::default();
+    if let Some(file) = file_layer {
+        // Deserialize the file layer on its own so unknown keys fail loudly
+        // with the file path attached, then fold into the merged struct.
+        let file_cfg = deserialize_layer_table(&file).map_err(|m| ConfigError::File {
+            path: file_path
+                .as_ref()
+                .map(|p| p.display().to_string())
+                .unwrap_or_default(),
+            message: m.to_string(),
+        })?;
+        fold(&mut raw, &file_cfg);
+    }
+    if !legacy_map.is_empty() {
+        let legacy_cfg = deserialize_layer_table(&legacy_map)?;
+        fold(&mut raw, &legacy_cfg);
+    }
+    if !env_map.is_empty() {
+        let env_cfg = deserialize_layer_table(&env_map)?;
+        fold(&mut raw, &env_cfg);
+    }
+
+    validate(&raw)?;
+    Ok(Loaded {
+        config: raw,
+        warnings,
+        file: file_path,
+    })
+}
+
+fn deserialize_layer_table(
+    map: &config::Map<String, config::Value>,
+) -> Result<DonsetchConfig, ConfigError> {
+    #[derive(Debug, Clone)]
+    struct MapSource(config::Map<String, config::Value>);
+    impl config::Source for MapSource {
+        fn collect(&self) -> Result<config::Map<String, config::Value>, config::ConfigError> {
+            Ok(self.0.clone())
+        }
+        fn clone_into_box(&self) -> Box<dyn config::Source + Send + Sync> {
+            Box::new(self.clone())
+        }
+    }
+    let builder = config::Config::builder().add_source(MapSource(map.clone()));
+    builder
+        .build()
+        .and_then(|c| c.try_deserialize::<DonsetchConfig>())
+        .map_err(|e| ConfigError::Env(e.to_string()))
+}
+
+/// Fold a layer into the accumulator: every field the layer explicitly
+/// set (differing from the struct default) is copied over the current
+/// value. Later layers win.
+fn fold(dst: &mut DonsetchConfig, src: &DonsetchConfig) {
+    dst.transport.fold_from(&src.transport);
+    dst.mcp.fold_from(&src.mcp);
+    dst.paths.fold_from(&src.paths);
+    dst.state.fold_from(&src.state);
+    dst.proxy.fold_from(&src.proxy);
+    dst.tls.fold_from(&src.tls);
+    dst.persona.fold_from(&src.persona);
+    dst.cli.fold_from(&src.cli);
+    dst.fetch.fold_from(&src.fetch);
+    dst.bypass.fold_from(&src.bypass);
+    dst.search.fold_from(&src.search);
+    dst.browser.fold_from(&src.browser);
+    dst.debug.fold_from(&src.debug);
+}
+
+/// The file layer path: `DONSETCH_CONFIG` explicit, else the default
+/// location when the file exists. `None` = no file layer.
+fn toml_path() -> Option<std::path::PathBuf> {
+    if let Some(explicit) = std::env::var_os("DONSETCH_CONFIG") {
+        return Some(std::path::PathBuf::from(explicit));
+    }
+    let dir = dirs::config_dir()?;
+    let path = dir.join("donsetch").join("donsetch.toml");
+    path.is_file().then_some(path)
+}
+
+// ---------------------------------------------------------------------------
+// Legacy env names
+// ---------------------------------------------------------------------------
+
+type VMap = config::Map<String, config::Value>;
+
+fn put(map: &mut VMap, path: &str, value: config::ValueKind, origin: &str) {
+    let mut parts = path.split('.');
+    let head = match parts.next() {
+        Some(p) => p,
+        None => return,
+    };
+    let origin = Some(origin.to_string());
+    let oref = origin.as_ref();
+    match parts.next() {
+        None => {
+            map.insert(head.to_string(), config::Value::new(oref, value));
+        }
+        Some(key) => {
+            let section = map
+                .entry(head.to_string())
+                .or_insert_with(|| config::Value::new(None, config::ValueKind::Table(VMap::new())));
+            if let config::ValueKind::Table(inner) = &mut section.kind {
+                inner.insert(key.to_string(), config::Value::new(oref, value));
+            }
+        }
+    }
+}
+
+/// Legacy strict opt-in flag: only `1`, `true`, `on` count.
+fn legacy_flag(name: &str) -> bool {
     env_flag_value(std::env::var_os(name).as_deref())
+}
+
+fn legacy_layer() -> (VMap, Vec<String>) {
+    let mut m = VMap::new();
+    let mut warnings = Vec::new();
+    let int_env = |name: &str| -> Option<i64> {
+        std::env::var(name)
+            .ok()
+            .and_then(|v| v.trim().parse::<i64>().ok())
+    };
+
+    // transport
+    if std::env::var("DONSETCH_TRANSPORT").as_deref() == Ok("http") {
+        put(
+            &mut m,
+            "transport.kind",
+            "http".into(),
+            "DONSETCH_TRANSPORT",
+        );
+    }
+    if let Some(v) = std::env::var_os("DONSETCH_HTTP_HOST") {
+        put(
+            &mut m,
+            "transport.host",
+            v.to_string_lossy().into_owned().into(),
+            "DONSETCH_HTTP_HOST",
+        );
+    }
+    match int_env("DONSETCH_HTTP_PORT") {
+        Some(p) => put(&mut m, "transport.port", p.into(), "DONSETCH_HTTP_PORT"),
+        None if std::env::var_os("DONSETCH_HTTP_PORT").is_some() => {
+            warnings.push("ignoring DONSETCH_HTTP_PORT: not a number".into())
+        }
+        None => {}
+    }
+    if let Some(v) = std::env::var_os("DONSETCH_HTTP_TOKEN") {
+        put(
+            &mut m,
+            "transport.token",
+            v.to_string_lossy().into_owned().into(),
+            "DONSETCH_HTTP_TOKEN",
+        );
+    }
+    match int_env("DONSETCH_HTTP_TIMEOUT_SECS") {
+        Some(s) => put(
+            &mut m,
+            "transport.timeout_secs",
+            s.into(),
+            "DONSETCH_HTTP_TIMEOUT_SECS",
+        ),
+        None if std::env::var_os("DONSETCH_HTTP_TIMEOUT_SECS").is_some() => {
+            warnings.push("ignoring DONSETCH_HTTP_TIMEOUT_SECS: not a number".into())
+        }
+        None => {}
+    }
+    if legacy_flag("DONSETCH_HTTP_CORS") {
+        put(&mut m, "transport.cors", true.into(), "DONSETCH_HTTP_CORS");
+    }
+
+    // mcp
+    if legacy_flag("DONSETCH_MCP_TEXT_ONLY") {
+        put(
+            &mut m,
+            "mcp.text_only",
+            true.into(),
+            "DONSETCH_MCP_TEXT_ONLY",
+        );
+    }
+    if legacy_flag("DONSETCH_NO_ANSWER_TOOL") {
+        put(
+            &mut m,
+            "mcp.answer_tool",
+            false.into(),
+            "DONSETCH_NO_ANSWER_TOOL",
+        );
+    }
+    if let Some(v) = std::env::var_os("DONSETCH_URL_HANDLES") {
+        // Historical semantics: every value except the exact string "off"
+        // keeps url handles enabled.
+        let on = v.to_str() != Some("off");
+        put(&mut m, "mcp.url_handles", on.into(), "DONSETCH_URL_HANDLES");
+    }
+
+    // state
+    if std::env::var_os("DONSEEK_NO_DISK_STATE").is_some() {
+        put(
+            &mut m,
+            "state.no_disk_state",
+            true.into(),
+            "DONSEEK_NO_DISK_STATE",
+        );
+    }
+    if legacy_flag("DONSETCH_NO_ROUTE_MEMORY") {
+        put(
+            &mut m,
+            "state.route_memory",
+            "off".into(),
+            "DONSETCH_NO_ROUTE_MEMORY",
+        );
+    }
+    if legacy_flag("DONSETCH_ROUTE_MEMORY_READONLY") {
+        put(
+            &mut m,
+            "state.route_memory",
+            "read_only".into(),
+            "DONSETCH_ROUTE_MEMORY_READONLY",
+        );
+    }
+    if legacy_flag("DONSETCH_NO_COOKIE_VAULT") {
+        put(
+            &mut m,
+            "state.cookie_vault",
+            false.into(),
+            "DONSETCH_NO_COOKIE_VAULT",
+        );
+    }
+    if std::env::var_os("DONSETCH_NO_WEB_MEMORY").is_some() {
+        put(
+            &mut m,
+            "state.web_memory",
+            false.into(),
+            "DONSETCH_NO_WEB_MEMORY",
+        );
+    }
+    match int_env("DONSETCH_WEB_MEMORY_CAP") {
+        Some(n) => put(
+            &mut m,
+            "state.web_memory_cap",
+            n.into(),
+            "DONSETCH_WEB_MEMORY_CAP",
+        ),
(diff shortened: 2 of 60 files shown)


Mart-Bogdan:
Hello, thanks, I'll take a look today.

I see that some tests are failing :-(.

Mart-Bogdan:
Hm it is windows breakage. I might try fixing it. but I am not fammiliar with that function probe_registry

Mart-Bogdan:
I have pushed fix of windows breakage. Tested it against real chrome and edge and detection in this function works.

Mart-Bogdan:
I have found one stray `env::var_os`  pushed fix. continuing review in read-only mode.

src/fetch/bypass.rs
```diff
@@ -65,6 +65,7 @@ impl Default for BypassConfig {
     }
 }
 
+#[allow(dead_code)]
 fn env_bool_off(name: &str) -> bool {
```
Mart-Bogdan: I think we should delete this function as it is no longer used and is dead code.

dondai44423:
Heads up on stack order: #198 (the web_answer + web_memory cut) is layered on top of this branch. Review this one first, #198 retargets to master after this merges.

src/ghost/cloak.rs
Mart-Bogdan, line 642: This test is now runner-dependent. `cfg()` freezes the config at the first call in the process, so the `set_var` / `remove_var` block only has an effect when this test is the first thing in a fresh process. Under nextest that holds and the assertion is real. Under plain `cargo test` whichever test touched `cfg()` first decides, and this passes vacuously. Same applies to every other test on the branch that mutates env and then calls converted code.

Rather than threading `&DonsetchConfig` through signatures, I would add a test-gated reset to `config.rs`. Keeps `cfg() -> &'static DonsetchConfig` and every call site as-is:

```rust
static CONFIG: std::sync::RwLock<Option<&'static DonsetchConfig>> =
    std::sync::RwLock::new(None);

pub fn cfg() -> &'static DonsetchConfig {
    if let Some(c) = *CONFIG.read().unwrap() {
        return c;
    }
    let mut w = CONFIG.write().unwrap();
    if let Some(c) = *w {
        return c;
    }
    let c: &'static DonsetchConfig = Box::leak(Box::new(load_or_default()));
    *w = Some(c);
    c
}

pub fn install(cfg: DonsetchConfig) -> Result<(), ConfigError> {
    validate(&cfg)?;
    let mut w = CONFIG.write().unwrap();
    if w.is_some() {
        return Err(ConfigError::Validate("config already initialized".into()));
    }
    *w = Some(Box::leak(Box::new(cfg)));
    Ok(())
}

/// Drop the frozen config so the next cfg() reloads from env + file.
/// Leaks the previous struct; test-only.
#[cfg(test)]
pub(crate) fn reset_for_test() {
    *CONFIG.write().unwrap() = None;
}
```

Production behavior is unchanged: one leaked struct for the process lifetime, an uncontended read lock per `cfg()` call. It also fixes `install()` silently ignoring an already-set config, which is a footgun on its own.

This test then becomes: mutate env, `crate::config::reset_for_test()`, resolve, restore env, reset again. Under nextest nothing changes. Under libtest the test now exercises the rule it claims to. If you would rather not touch the global at all, the alternative is a pure `config_backend_for(&DonsetchConfig)` sibling and a test that builds the struct directly, the way `cap_for` in `memory/store.rs` already does.

src/fetch/client.rs
```diff
@@ -267,8 +267,7 @@ impl Fetcher {
         // and crawl (many pages, same host) where rate limits bite.
 
         loop {
-            let env_proxy = if proxy.is_none() && !crate::config::env_flag("DONSETCH_NO_ENV_PROXY")
-            {
+            let env_proxy = if proxy.is_none() && crate::config::cfg().proxy.from_environment {
```
Mart-Bogdan: `proxy.from_environment = false` gates the whole `from_env_for()` call, but the explicit `proxy.https` / `proxy.http` / `proxy.all` slots are resolved *inside* that function. So `from_environment = false` plus a TOML-configured proxy yields no proxy at all. The switch was meant to drop the ambient layer only. Suggest: always call `from_env_for`, and have it skip the `HTTPS_PROXY`/`ALL_PROXY` fallback when `from_environment` is false. Same gate at `doctor.rs:306`.

src/cli/doctor.rs
```diff
@@ -283,7 +303,7 @@ async fn check_network(fetcher: &Fetcher) -> CheckResult {
 /// and the two certificate stores the connector builds from.
 /// Local-only: no network, stays in fast mode.
 fn check_fetch_egress() -> CheckResult {
-    let kill = crate::config::env_flag("DONSETCH_NO_ENV_PROXY");
+    let kill = !crate::config::cfg().proxy.from_environment;
```
Mart-Bogdan: Same issue as `client.rs:270`: with `from_environment = false` this reports "direct egress" even when the TOML configures a proxy, and the daemon then fetches direct too.

src/cli/doctor.rs
```diff
@@ -297,7 +317,8 @@ fn check_fetch_egress() -> CheckResult {
         bits.push(format!("egress via {} proxy {}:{} (env; SOCKS5 keeps TLS end-to-end, HTTP CONNECT gets the interception-safe handshake)", if p.is_http_connect() { "http" } else { "socks5" }, p.host, p.port));
     } else {
         bits.push(if kill {
-            "direct egress (DONSETCH_NO_ENV_PROXY=1 disables the env-proxy convention)".into()
+            "direct egress (proxy.from_environment = false disables the env-proxy convention)"
```
Mart-Bogdan: The hints in this function and in `check_network` (L290, L313) look only at the ambient `HTTPS_PROXY` / `SSL_CERT_FILE`. A TOML-configured `proxy.https` or `tls.cert_file` gets the "no proxy env vars, export HTTPS_PROXY" advice. Cosmetic, but doctor is where people go when egress fails; reading `cfg().proxy` / `cfg().tls` first would keep the advice honest.

src/config.rs
```diff
@@ -1,17 +1,2089 @@
...
+    // browser
+    if let Some(v) = std::env::var_os("DONSETCH_BROWSER_BACKEND")
+        .or_else(|| std::env::var_os("DONGHOST_BROWSER_BACKEND"))
+    {
+        let v = v.to_string_lossy().to_ascii_lowercase();
+        let backend = match v.as_str() {
+            "auto" | "" => "auto",
+            "chromium" | "chrome" | "original" => "chromium",
+            "cloak" | "cloakbrowser" => "cloak",
+            _ => {
```
Mart-Bogdan, line 913: DONSETCH_BROWSER_BACKEND=headless is silently dropped. The legacy mapper knows only auto, chromium, cloak. headless and original-headless hit the warning branch although the enum has the variant and the README documents the value.

src/config.rs
```diff
@@ -1,17 +1,2089 @@
...
+        if std::env::var_os(&env_name).is_some() {
+            origin = env_name;
+        } else {
+            for var in LEGACY_VARS {
+                if legacy_target_of(var) == (section, key) && std::env::var_os(var).is_some() {
+                    origin = format!("legacy:{var}");
+                    break;
+                }
+            }
+            if origin == "default"
```
Mart-Bogdan: Provenance is wrong whenever a TOML file exists: this branch stamps the file as the origin of every knob still at its default, whether or not the file mentions that key. A one-line `donsetch.toml` makes all ~70 rows of `config show` say `file:...`.

The information already exists: `load()` collects the file layer into a `config::Map` before folding it, and that map holds exactly the keys the file set. Suggested fix:

1. Have `load()` walk that map once and record the `(section, key)` pairs it contains:

```rust
fn keys_of(map: &VMap) -> std::collections::BTreeSet<(String, String)> {
    let mut out = BTreeSet::new();
    for (section, v) in map {
        if let config::ValueKind::Table(inner) = &v.kind {
            for key in inner.keys() {
                out.insert((section.clone(), key.clone()));
            }
        }
    }
    out
}
```

and return them on `Loaded`, next to `file`: `pub file_keys: BTreeSet<(String, String)>`. The legacy map can get the same treatment (`legacy_keys`), which also stops a legacy var that was *ignored* with a warning from being reported as the origin.

2. Make `origins()` take `&Loaded` instead of `&DonsetchConfig` (the only caller is `show_text`, called once from `main.rs` with a `Loaded` in hand), and resolve per knob:

```rust
let origin = if std::env::var_os(&env_name).is_some() {
    env_name
} else if let Some(var) = legacy_setter(section, key, &loaded.legacy_keys) {
    format!("legacy:{var}")
} else if loaded.file_keys.contains(&(section.into(), key.into())) {
    format!("file:{}", loaded.file.as_ref().unwrap().display())
} else {
    "default".into()
};
```

3. Keep the branch order here identical to the fold order in `load()`. Right now the doc comment says "new env beats legacy beats file", which matches `load()` but not the module doc / README; whichever way that gets settled, this function and `fold()` must agree, otherwise `config show` reports a setter that lost.

A unit test that builds a one-key file map, runs `origins()`, and asserts exactly one row is `file:` and the rest `default` would pin it.

dondai44423:
Thanks both, real catches.

Mart-Bogdan: verified both your pushes. The probe_registry breakage was mine, the config pass mangled that function, your sort-by-configured-family version is the right shape and the env::var_os straggler removal is correct. Tested against the tree, good.

itoqa: Good job, finally of some use

Gate on the branch head after the fix: 1155 passed + 1 skipped, green twice.

src/config.rs
```diff
@@ -1,17 +1,2089 @@
...
+    let (legacy_map, legacy_warnings) = legacy_layer();
+    warnings.extend(legacy_warnings);
+    let (env_map, env_warnings, env_errors) = new_env_layer();
+    warnings.extend(env_warnings);
+    if let Some(e) = env_errors.first() {
+        return Err(ConfigError::Env(e.clone()));
+    }
+
+    let mut raw = DonsetchConfig::default();
+    if let Some(file) = file_layer {
```
Mart-Bogdan: doc comment at top of module says:
```
//! Layers, lowest to highest precedence:
//!   1. serde defaults (this file, matching the historical runtime defaults)
//!   2. legacy env names (the pre-v4 scattered knobs, exact trigger semantics preserved)
//!   3. `donsetch.toml` file layer (`DONSETCH_CONFIG` explicit, or the default location)
//!   4. new env names: `DONSETCH_<SECTION>__<KEY>` (two underscores = one dot)
```

so for this to work this way we need to swap 329-340 and 341-344

One consequence worth mentioning explicitly: with legacy below the file, a user who still has `DONSETCH_PDF_MAX_MB=10` exported and writes `pdf_max_mb = 50` in a deployment TOML gets 50. That is the intended per-harness override, but it is the opposite of what the current code does, so it should be a conscious choice in the commit message rather than a silent flip.

Also `donsetch::config::origins` (L1864) must be aligned with this. It encodes the same priority as the current fold order: when a legacy var and the TOML both set a knob, it reports the legacy var as the source, and its comment says "legacy names beat the file layer". After swapping the blocks here, that function needs the opposite priority, file over legacy, or `config show` will name a setter that lost.

src/config.rs
```diff
@@ -1,17 +1,2089 @@
...
+            fn default() -> Self {
+                Self { $($field: $default,)* }
+            }
+        }
+        impl $name {
+            /// Fold one layer into the accumulator: copy every field the
+            /// layer explicitly set (differing from the struct default).
+            pub(crate) fn fold_from(&mut self, other: &Self) {
+                let defaults = Self::default();
+                $( if other.$field != defaults.$field {
```
Mart-Bogdan: A higher layer cannot set a value back to its default. `fold_from` copies a field only when it differs from the struct default, so an explicit "set to the default value" is indistinguishable from "not set". Examples: TOML `answer_tool = false` plus `DONSETCH_MCP__ANSWER_TOOL=true` stays false; TOML `h3 = true` plus `DONSETCH_FETCH__H3=false` stays true; a legacy `DONSETCH_NO_WEB_MEMORY` cannot be undone by `web_memory = true` in a deployment TOML. That defeats the per-harness override case from #193.

Fix options: (a) merge the raw `config::Map` layers before a single `try_deserialize`, so presence is tracked by the map rather than inferred from values (the `config` crate's builder does this natively if the layers are added as sources in order); or (b) make each layer's section structs `Option`-typed and fold on `Some`. (a) is less code and also gives `origins()` its provenance for free.

Side effect, verified on the built binary: `origins()` decides provenance from "is the env var present", not from what the fold kept, so in this exact case `config show` prints `answer_tool = false (DONSETCH_MCP__ANSWER_TOOL)`. The value and the named source contradict each other, and a user debugging this sees their own variable blamed for a value they did not set. Fixing the fold by merging the raw maps (option a) also makes the origin truthful, since provenance can then be read from the same merged map.

src/config.rs
```diff
@@ -1,17 +1,2089 @@
...
+        }
+    }
+}
+
+/// Legacy strict opt-in flag: only `1`, `true`, `on` count.
+fn legacy_flag(name: &str) -> bool {
     env_flag_value(std::env::var_os(name).as_deref())
 }
 
+fn legacy_layer() -> (VMap, Vec<String>) {
```
Mart-Bogdan, line 455: Legacy values are no longer preprocessd/clamped. This map is deserialized at L342 and the merged struct goes through `validate()` at L350, so legacy values that the old code clamped or silently ignored now make `load()` fail, which is fatal for `donsetch mcp`. `int_env` already downgrades *unparseable* values to a warning; the regression is for values that parse but fall outside the new ranges, plus every negative integer (the `config` crate maps `i64 -> u64` via `try_into` and errors on negatives, so those fail at L342 before validation even runs).

| Legacy var | Old behavior on master | New behavior |
|---|---|---|
| `DONSETCH_WEB_MEMORY_CAP=100` | ignored, default 4000 (`memory/store.rs:57`) | `validate()` rejects `<256`, `load()` fails |
| `DONSETCH_BYPASS_CACHE_TTL_SECS=604800` | accepted, clamp was `0..=31_536_000` (`fetch/bypass.rs:110`) | rejects `>86_400` |
| `DONSETCH_GHOST_POOL_SLOTS=99` | clamped to 16 (`ghost/manager.rs:53`) | rejects `>16` |
| `DONSETCH_XVFB_DISPLAY=300` | fell back to 99 (`ghost/xvfb.rs:33`) | rejects `>254` |
| `DONSEEK_RERANK_THREADS=128` | valid, max was `i32::MAX` (`search/rerank.rs:38`) | rejects `>64` |
| any negative, e.g. `DONSETCH_PDF_MAX_MB=-1` | parse failed, default used (`pdf/mod.rs:119`) | type error in `try_deserialize`, before validation |

This contradicts the "exact historical trigger semantics preserved" promise that justified keeping the shim, and the `DONSEEK_RERANK_THREADS` graceful exception from the #193 thread. Suggested: apply the historical clamp / fallback inside this function (warn on stderr, then clamp or drop the key), so only the TOML and `DONSETCH_*` layers are strict.

Mart-Bogdan (changes requested):
Thanks. This is the shape #193 asked for. The field book, `config show` and `--legacy` are exactly the self-documenting part I hoped for.

The blocking items are in inline threads: fold-by-difference, layer order vs. docs, legacy values failing validation, `from_environment = false` dropping a TOML proxy too, `origins()` provenance, `headless` dropped by the legacy mapper.

## Smoke test on the built binary

Debug build of `131eb0a` on Windows 11, run with `DONSETCH_CACHE_DIR` pointed at a scratch dir. The load, `config show`, `--legacy`, `--markdown`, the `DONSETCH_CONFIG` override and every error path work as described in the PR. Each inline finding also reproduces:

| Finding (inline thread) | What I ran | What happened |
|---|---|---|
| Fold-by-difference (`config.rs:52`) | TOML `answer_tool = false` + `DONSETCH_MCP__ANSWER_TOOL=true` | `config show` prints `answer_tool = false (DONSETCH_MCP__ANSWER_TOOL)`. The env value was dropped, and the origin column names the env var anyway. |
| Layer order (`config.rs:329`) | TOML `pdf_max_mb = 50` + legacy `DONSETCH_PDF_MAX_MB=10` | value is `10`. The legacy var beat the file, opposite to the module doc. |
| Provenance (`config.rs:1878`) | a TOML with one key, `[mcp] answer_tool = false` | 72 of 74 rows show `(file:...)` as their origin, although the file sets only one key. |
| Legacy now fatal (`config.rs:442`) | `DONSETCH_WEB_MEMORY_CAP=100`; separately `DONSETCH_PDF_MAX_MB=-1` | both exit 1. The first fails validation, the second fails with a type error naming the var before validation runs. |
| `headless` dropped (`config.rs:840`) | `DONSETCH_BROWSER_BACKEND=headless` | warning "ignoring unknown browser backend", backend stays `auto`. |
| Unknown var warning (notes below) | `DONSETCH_DEBUG=1` | `ignoring DONSETCH_DEBUG: expected DONSETCH_<SECTION>__<KEY>` on every run. |
| `NO_CONFIG_FILE=0` (notes below) | `DONSETCH_NO_CONFIG_FILE=0` + `DONSETCH_CONFIG=<file>` | errors with "NO_CONFIG_FILE=1 but DONSETCH_CONFIG is set". The value `0` is treated as set. |
| Non-mcp commands continue on defaults (notes below) | `donsetch status` with a TOML that has an unknown key | prints the error, then "using defaults", then runs normally and exits 0. `config show` with the same file exits 1. |

Two more things below. They could not go inline because the lines are not part of the diff.

## Env reads that were not converted

These files still read the old env var directly. The new TOML / `DONSETCH_*` key for the same knob does nothing there. Only the legacy name works.

| Raw read | Key it bypasses | Effect |
|---|---|---|
| [`adapters/mod.rs:30`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/adapters/mod.rs#L30) `DONSETCH_NO_ADAPTERS` | `fetch.adapters` | key does nothing; only the legacy var disables adapters |
| [`adapters/mod.rs:37`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/adapters/mod.rs#L37) `DONSETCH_ADAPTER_DUMP` | `fetch.adapter_dump_dir` | same |
| [`cli/adapters.rs:12`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/cli/adapters.rs#L12) `DONSETCH_NO_ADAPTERS` | `fetch.adapters` | CLI listing disagrees with the daemon |
| [`search/engines.rs:171`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/engines.rs#L171), [`byok/mod.rs:243`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/mod.rs#L243), [`byok/mod.rs:275`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/mod.rs#L275), [`byok/plugin.rs:409`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/plugin.rs#L409), [`byok/plugin.rs:442`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/plugin.rs#L442), [`byok/plugin.rs:544`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/plugin.rs#L544) `DONSEEK_DEBUG` | `debug.search` | six search debug lines ignore the key |
| [`pdf/rotate.rs:43`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/pdf/rotate.rs#L43) `DONSHEET_DEBUG_MATRIX`, [`pdf/rotate.rs:55`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/pdf/rotate.rs#L55) `DONSHEET_DEBUG` | `debug.pdf_matrix`, `debug.pdf` | rotation debug ignores the keys |

Eleven sites. The adapters pair changes behavior. The rest only affect debug output. Found with:

```sh
rg -n 'env::(var|var_os)\s*\(\s*"(DON[A-Z]+_|CLOAKBROWSER_)' src --glob '!config.rs'
```

That command returns 15 hits. The other four are intentional: `paths.rs:20` reads `DONSETCH_CACHE_DIR` per call on purpose, and `ghost/cloak.rs:650-652` is inside a test.

Related: `browser.playwright_path` is in the field book, but no code reads it. [`ghost/mod.rs:313`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/ghost/mod.rs#L313) still reads `PLAYWRIGHT_BROWSERS_PATH` directly. Either wire the field or remove the entry.

## Smaller notes

- Only `mcp` and `config show` fail on a bad config. `donsetch fetch` with a broken TOML prints "using defaults" and continues (`config.rs:2064`). The PR description says bad values are hard errors.
- `install()` does nothing if the `OnceLock` is already set (`config.rs:2046`). Today nothing calls `cfg()` before it in the `mcp` path. But one log line added before it would silently drop `--port`. It should return an error.
- `DONSETCH_NO_GHOST_POOL` together with `DONSETCH_GHOST_POOL_SLOTS`: the old rule was "kill switch wins". Now the later `put` wins, so the explicit count does (`config.rs:904`).
- `DONSETCH_NO_CONFIG_FILE` only checks presence (`config.rs:289`), so `=0` still skips the file. Every other boolean in the new env layer goes through `boolish()` and rejects values that are not true/false words. This one should too.
- Every unknown `DONSETCH_*` var prints a warning at start (`config.rs:1731`). The pi extension sets `DONSETCH_DEBUG` for its own logging, so every pi launch now gets that warning. Either add it to `RESERVED_VARS`, or rename it on the pi side to something outside the `DONSETCH_` prefix, since any name in that prefix without `__` warns.
- `Box::leak` in [`search/rerank.rs:150`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/rerank.rs#L150) runs once per session build. Harmless. An owned enum variant would avoid it.
- BYOK plugins do not get a proxy that is set only in TOML. Chromium gets it through `from_env_for`. Plugin children only receive `DONSETCH_PLUGIN*` ([`search/byok/plugin.rs:599`](https://github.com/dondai44423/donsetch/blob/config-consolidation/src/search/byok/plugin.rs#L599)).
- `proxy.no_proxy` is a string, `proxy.pool` is a list. The field book calls `no_proxy` a comma list. A list would match the proposal.
- `tls.cert_file` / `tls.cert_dir` do not replace `SSL_CERT_FILE` / `SSL_CERT_DIR`. `env_root_paths()` loads both sets (`transport/tls.rs:307`). Additive is right for trust roots. The field book should say so, since every other mirrored knob overrides its ambient var.

---

To pull all inline threads of this review as plain text (`gh pr view --comments` shows only top-level comments, not inline ones):

```sh
gh api repos/dondai44423/donsetch/pulls/196/comments --paginate \
  --jq '.[] | "=== \(.path):\(.line // .original_line)\n\(.body)\n"'
```

Mart-Bogdan:
@dondai44423 I have noticed two Windows breakages today. What OS are you using on your box, Linux or macOS?

dondai44423:
@Mart-Bogdan  I use linux (void)

src/config.rs
Mart-Bogdan: Here is PoC example:
<details>
  <summary>main.rs</summary>

```rs
//! PoC: let the `config` crate do the layer merge, then read the winning
//! origin per key from the merged tree. No fold_from, no env re-probing.

use config::{Map, Source, Value, ValueKind};

#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
#[serde(default, deny_unknown_fields)]
struct Mcp {
    text_only: bool,
    answer_tool: bool,
}

#[derive(Debug, serde::Deserialize, serde::Serialize)]
#[serde(default, deny_unknown_fields)]
struct Fetch {
    h3: bool,
    pdf_max_mb: u64,
}
impl Default for Fetch {
    fn default() -> Self {
        Self { h3: false, pdf_max_mb: 100 }
    }
}

#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
#[serde(default, deny_unknown_fields)]
struct Cfg {
    mcp: Mcp,
    fetch: Fetch,
}

/// Same shape as the PR's `put()`: nested table, origin stamped on the leaf.
fn put(map: &mut Map<String, Value>, path: &str, v: ValueKind, origin: &str) {
    let (sec, key) = path.split_once('.').unwrap();
    let o = Some(origin.to_string());
    let section = map
        .entry(sec.into())
        .or_insert_with(|| Value::new(None, ValueKind::Table(Map::new())));
    if let ValueKind::Table(inner) = &mut section.kind {
        inner.insert(key.into(), Value::new(o.as_ref(), v));
    }
}

#[derive(Debug, Clone)]
struct MapSource(Map<String, Value>);
impl Source for MapSource {
    fn collect(&self) -> Result<Map<String, Value>, config::ConfigError> {
        Ok(self.0.clone())
    }
    fn clone_into_box(&self) -> Box<dyn Source + Send + Sync> {
        Box::new(self.clone())
    }
}

/// Re-create every leaf with the given origin (config::Value.origin is private).
fn restamp(m: Map<String, Value>, origin: &str) -> Map<String, Value> {
    let o = Some(origin.to_string());
    m.into_iter()
        .map(|(k, v)| {
            let kind = match v.kind {
                ValueKind::Table(t) => ValueKind::Table(restamp(t, origin)),
                other => other,
            };
            (k, Value::new(o.as_ref(), kind))
        })
        .collect()
}

fn walk(prefix: &str, m: &Map<String, Value>, out: &mut Vec<(String, String, String)>) {
    for (k, v) in m {
        let path = if prefix.is_empty() { k.clone() } else { format!("{prefix}.{k}") };
        match &v.kind {
            ValueKind::Table(t) => walk(&path, t, out),
            other => out.push((
                path,
                format!("{other}"),
                v.origin().unwrap_or("default").to_string(),
            )),
        }
    }
}

fn main() {
    // Layer 1: defaults, from the struct (origin = None -> "default").
    let defaults = config::Config::try_from(&Cfg::default()).unwrap();

    // Layer 2: legacy env (simulated). DONSETCH_NO_ANSWER_TOOL -> answer_tool=false.
    let mut legacy = Map::new();
    put(&mut legacy, "mcp.answer_tool", false.into(), "legacy:DONSETCH_NO_ANSWER_TOOL");
    put(&mut legacy, "fetch.pdf_max_mb", 10i64.into(), "legacy:DONSETCH_PDF_MAX_MB");

    // Layer 3: TOML file. Sets pdf_max_mb=50 and h3=true.
    let toml_text = "[fetch]\npdf_max_mb = 50\nh3 = true\n";
    let mode = std::env::args().nth(1).unwrap_or_default();
    let file: Box<dyn Source + Send + Sync> = match mode.as_str() {
        // A: path-based source; config stamps the file path as origin itself.
        "path" => {
            let p = std::env::temp_dir().join("cfgpoc.toml");
            std::fs::write(&p, toml_text).unwrap();
            Box::new(config::File::from(p))
        }
        // B: read-to-string as the PR does, then stamp origins on the collected map.
        "stamp" => {
            let m = config::File::from_str(toml_text, config::FileFormat::Toml).collect().unwrap();
            Box::new(MapSource(restamp(m, "file:/etc/donsetch.toml")))
        }
        _ => Box::new(config::File::from_str(toml_text, config::FileFormat::Toml)),
    };

    // Layer 4: new env (simulated). DONSETCH_MCP__ANSWER_TOOL=true -> back to the DEFAULT value.
    let mut env = Map::new();
    put(&mut env, "mcp.answer_tool", true.into(), "DONSETCH_MCP__ANSWER_TOOL");

    let merged = config::Config::builder()
        .add_source(defaults)
        .add_source(MapSource(legacy))
        .add_source(vec![file])
        .add_source(MapSource(env))
        .build()
        .unwrap();

    // Provenance: walk the merged tree, each leaf keeps the origin of the layer that won.
    let mut rows = Vec::new();
    walk("", &merged.collect().unwrap(), &mut rows);
    rows.sort();
    println!("key                value   origin");
    for (k, v, o) in &rows {
        println!("{k:<18} {v:<7} {o}");
    }

    // Typed struct: one deserialize at the end.
    let cfg: Cfg = merged.try_deserialize().unwrap();
    println!("\n{cfg:#?}");

    // Checks the PR's fold_from gets wrong.
    assert!(cfg.mcp.answer_tool, "env must be able to set a value back to its default");
    assert_eq!(cfg.fetch.pdf_max_mb, 50, "file must beat legacy");
    assert!(cfg.fetch.h3);
    println!("\nOK: explicit-default override works, file beats legacy, origins per key.");

    // deny_unknown_fields still fires on a bad file key.
    let bad = config::Config::builder()
        .add_source(config::Config::try_from(&Cfg::default()).unwrap())
        .add_source(config::File::from_str("[fetch]\nnope = 1\n", config::FileFormat::Toml))
        .build()
        .unwrap()
        .try_deserialize::<Cfg>();
    println!("bad key -> {}", bad.err().map(|e| e.to_string()).unwrap_or("NO ERROR".into()));
}
```

</details>

<details>
  <summary>Cargo.toml</summary>

```toml
[package]
name = "cfgpoc"
version = "0.1.0"
edition = "2021"

[dependencies]
config = { version = "0.15", default-features = false, features = ["toml"] }
serde = { version = "1", features = ["derive"] }

[workspace]
```

</details>

<details>
  <summary>Output</summary>
  
```
key                value   origin
fetch.h3           true    default
fetch.pdf_max_mb   50      default
mcp.answer_tool    true    DONSETCH_MCP__ANSWER_TOOL
mcp.text_only      false   default

Cfg {
    mcp: Mcp {
        text_only: false,
        answer_tool: true,
    },
    fetch: Fetch {
        h3: true,
        pdf_max_mb: 50,
    },
}

OK: explicit-default override works, file beats legacy, origins per key.
bad key -> unknown field `nope`, expected `h3` or `pdf_max_mb` for key `fetch`

```

</details>

Mart-Bogdan: Note on the output above: `fetch.h3` and `fetch.pdf_max_mb` show `default` because that run used `File::from_str`, which carries no URI, so the leaves come out with no origin. That is exactly what `load()` does today after reading the file itself. Two fixes, both verified in the PoC (`cfgpoc path` and `cfgpoc stamp`):

- `config::File::from(path)` instead of `from_str`. The crate stamps the path as origin.
- Keep `from_str`, `collect()` its map, and rebuild each leaf with `Value::new(Some("file:<path>"), kind)`. The `origin` field is private, hence the `restamp()` helper.

With either, the same run prints:

```
key                value   origin
fetch.h3           true    file:/etc/donsetch.toml
fetch.pdf_max_mb   50      file:/etc/donsetch.toml
mcp.answer_tool    true    DONSETCH_MCP__ANSWER_TOOL
mcp.text_only      false   default
```

Change size in `load()`: build the four maps as now, add them to the builder in order, one `try_deserialize`, then `validate()`. `fold`, the thirteen `fold_from` methods and the macro's fold arm go away; `origins()` becomes the walk over `merged.collect()` joined with the field book. The per-file error attribution survives by deserializing the file map alone as a validation step before adding it.

Mart-Bogdan:
I have just added new comment under one of inline comments, regarding proposed code change, in case it would be missed in notifications: https://github.com/dondai44423/donsetch/pull/196#discussion_r3992728123

dondai44423:
Mart-Bogdan: your proposed design is better than what I had, I adopted it wholesale. The layering now rides the config crate builder (defaults < legacy < file < DONSETCH_ env), so presence is preserved by the builder and each merged leaf carries the origin of the layer that won it. The hand-rolled fold is deleted, and origins() reads straight off the merged tree.

Everything from the review is fixed on the branch head (b8c2f2a):

- Presence-based layering: a top layer setting a field back to its default still wins and owns the origin.
- Legacy numerics keep soft semantics: out-of-range warns and keeps the default, never a hard exit. headless backend spelling restored, ghost-pool kill switch beats the slot count again, falsy NO_CONFIG_FILE words still load the file.
- proxy.from_environment now gates only the ambient convention: explicit TOML slots stay live and plugin child envs carry them.
- The file origin only labels keys actually present in the file.
- Non-mcp commands hard-fail on a broken config instead of running on defaults.
- playwright_path, the adapters pair and the debug toggles wired to the typed config; DONSETCH_DEBUG reserved.
- install() errors on a double install.

8 discriminating tests pin each finding, one per repro from your review. Gate on the head: 1163 passed + 1 skipped, green twice. Your Windows fixes are verified and stay in.

Mart-Bogdan:
Thanks for the fast turnaround, and glad the builder-based layering worked out. I won't get to it today, but I'll go through the new head (ae84f6b913) and re-run my checks tomorrow.

dondai44423:
@itoqa good catch, that one was real. The wave installed the config once for every command, and the mcp arm still ran its own install: the second install hit the one-shot check and the daemon exited before serving. Fixed on the head (3d41764): the arm reads the installed config. A boot contract test drives the real binary over stdio, pins the initialize answer, and fails on the pre-fix signature (the double-install stderr line) while passing on the fix.

src/config.rs
```diff
@@ -1,17 +1,2171 @@
...
+        return err("search.rerank_threads must be 0 (auto) or 1..=64".into());
+    }
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Show: origins, human table, markdown table
+// ---------------------------------------------------------------------------
+
+fn value_of(c: &DonsetchConfig, section: &str, key: &str) -> String {
```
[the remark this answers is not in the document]
Mart-Bogdan: Good finding, probable fix is to create helper `is_secret(section, key)` and use it in `show_text` and either replace whole secret with ****  or keep couple chars from frond and back and replace rest with ****

tests/daemon_boot_stays_alive.rs
```diff
@@ -0,0 +1,63 @@
...
+//! first byte. This test drives the real binary over stdio and pins
+//! both the response and the absence of the double-install error.
+
+use std::io::{BufRead, BufReader, Write};
+use std::process::{Command, Stdio};
+use std::sync::mpsc;
+use std::time::Duration;
+
+#[test]
+fn mcp_daemon_boots_and_answers_initialize() {
```
Mart-Bogdan: This spawns the real binary with the developer's ambient env, so it also picks up a real `donsetch.toml` if one exists. With `[transport] kind = "http"` in that file the child starts the HTTP server, never answers on stdio, and the test fails on the 20 s timeout for reasons unrelated to the double-install regression. Isolating the child fixes it:

```rust
.env("DONSETCH_NO_CONFIG_FILE", "1")
```

Same for any `DONSETCH_TRANSPORT=http` in the shell, so an `env_remove` for that one is worth adding too.

Smaller: the stderr check reads only the first line. Any `[donsetch config]` warning (a stray legacy var in the developer's shell) lands there first and the `already installed` assertion becomes vacuous. Reading stderr to EOF before asserting keeps it meaningful. The initialize assertion still catches the regression either way, so this is a robustness nit, not a correctness one.

src/cli/doctor.rs
Mart-Bogdan: @dondai44423 you probably missed this comment, proxy.rs `from_env_for` now reads the TOML slots first and applies `from_environment` only to the ambient HTTPS_PROXY fallback. 

So doctor must call `proxy::from_env_for` and do not check `proxy.from_environment` beforehand, as `from_env_for` already honors the TOML slots and gates only the ambient env read.

Proposed fix:
```rs
let resolved = crate::transport::proxy::from_env_for("https://example.com");
...
} else {
    bits.push(if crate::config::cfg().proxy.from_environment {
        "direct egress (no proxy env vars; export HTTPS_PROXY/HTTP_PROXY to route fetches)"
    } else {
        "direct egress (proxy.from_environment = false disables the env-proxy convention)"
    }.into());
}
```

src/cli/doctor.rs
```diff
@@ -297,7 +317,8 @@ fn check_fetch_egress() -> CheckResult {
         bits.push(format!("egress via {} proxy {}:{} (env; SOCKS5 keeps TLS end-to-end, HTTP CONNECT gets the interception-safe handshake)", if p.is_http_connect() { "http" } else { "socks5" }, p.host, p.port));
```
Mart-Bogdan: Nit:
The success line still says (env; SOCKS5 ...) for a proxy that came from TOML. Proxy carries no origin, so "config" vs "env" would need a check on `cfg().proxy.https`/`http`/`all` being non-empty, or just drop the word "env".

src/ghost/cloak.rs
Mart-Bogdan: Not addressed in the revision, but I can live with that: the suite is nextest-only anyway, and under process-per-test the `set_var` block runs before anything touches `cfg()`, so the assertion is real there. Under plain `cargo test` it is still vacuous. If you want to skip the reset helper, please add a short comment on the test saying it relies on nextest process isolation and that `cfg()` freezes at first call, so the next person who runs it under libtest knows why it passes for the wrong reason. Same note applies to `from_environment_off_keeps_toml_proxy_slots_alive` in `config.rs`, which also reaches `cfg()` lazily.

src/cli/doctor.rs
Mart-Bogdan: Still open in the revision. Same fix as the two threads above: read `cfg().proxy` and `cfg().tls` before falling back to the ambient vars, so the advice matches what the daemon will actually do.

src/fetch/bypass.rs
Mart-Bogdan: Nit:
This dead code should probably be removed, but I don't insist on it.

Mart-Bogdan (approved):
Re-reviewed on `3d41764`. All four blocking items from the first round are fixed and pinned by tests: builder layering with presence semantics, legacy below the file, soft-range legacy ints, `from_environment` gating only the ambient read. I resolved those threads (You can scroll this PR discussion on github "Conversation" tab and see what threads aren't still resolved).

Still open: the **itoqa** finding that `config show` prints `transport.token` and proxy userinfo verbatim (new with this PR, needs a redaction pass before merge), the `doctor` half of the proxy fix, the two `doctor` hint threads, and two small nits (dead `env_bool_off`, a comment on the nextest-dependent tests). Details inline.

Approving from my side, but the **itoqa** finding on `config show` should still be addressed before merge.

src/config.rs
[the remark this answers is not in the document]
dondai44423: Done on d833e47.

tests/daemon_boot_stays_alive.rs
dondai44423: Done on d833e47.

src/cli/doctor.rs
dondai44423: Done on d833e47.

src/cli/doctor.rs
dondai44423: Done on d833e47.

src/ghost/cloak.rs
dondai44423: Done on d833e47.

src/cli/doctor.rs
dondai44423: Done on d833e47.

src/fetch/bypass.rs
dondai44423: Done on d833e47.

dondai44423:
All inline findings fixed on d833e47 (gate x2 green, 1174 + 1 skip):

- config show masks the credential-bearing proxy fields (https/http/all/pool): scheme://***@host, pool masked per entry. Covers the bot's high finding from both sides.
- check_fetch_egress calls from_env_for unconditionally now (it already consults the [proxy] slots first and gates only the ambient read), labels the winning layer config vs env, and the no-proxy message says the slots stay live.
- The failed-direct-fetch hint fires on a proxy configured by either layer and names both fixes ([proxy] slots, [tls] cert_file).
- The boot contract test is hermetic (DONSETCH_NO_CONFIG_FILE=1, ambient DONSETCH_* stripped) and reads stderr to EOF.
- Isolation notes added to the two env-mutating tests, one dead helper dropped from bypass.rs.

@itoqa the config show leak is fixed in the same commit; the daemon boot finding from your earlier run stays fixed (your rerun already verified it).

src/config.rs
```diff
@@ -1965,6 +1965,56 @@ pub fn origins(
...
+    if url.is_empty() {
+        String::new()
+    } else {
+        "***".to_string()
+    }
+}
+
+/// `config show` display form for one knob: value_of output for
+/// everything, credentials masked for the secret proxy fields.
+fn masked_value(c: &DonsetchConfig, section: &str, key: &str, shown: String) -> String {
```
Mart-Bogdan, line 1996: Proxy masking looks right, but `transport.token` still prints in clear: `is_secret` only matches the proxy keys, and the bearer token was the first item in the finding. One more arm for `("transport", "token")` printing `"***"`, plus an assert in `show_text_redacts_proxy_secrets`, closes it.

Mart-Bogdan (commented):
Checked `d833e47`: doctor proxy/hint threads, the boot test isolation, the test notes and the dead helper are all done. One gap left from the itoqa finding, inline.

adaaaaaaaaaaaaaaaaaaaaaa:
I prepared #203 as a focused follow-up to `d833e47` for the remaining parts of the Itoqa credential-disclosure finding.

It redacts `transport.token` and masks proxy credentials using the final `@`, so passwords containing `@` cannot remain partially visible. The regression test covers every proxy slot, pools, bare values, empty values, retained non-secret diagnostics, and the transport token.

Verified with the 19 config tests, clippy with warnings denied, formatting checks, and a live `donsetch config show` exercise.

PR #197, opened 2026-09-11
fix(search): cache BYOK results so repeat queries stop re-billing the provider (#195)

### LINKED ISSUE #195: BYOK results are never cached (repeated identical queries keep billing the provider)
## Environment
- donsetch **3.6.7**, macOS arm64, BYOK provider `tinyfish` (reproduced identically with `serper`)

## Summary
The local path caches search results and reports `meta.cached = true` on a repeat.
The **BYOK path never reads nor writes that cache** — `meta.cached` is `false` on every call, and no entry is written to `search-cache.json`.

## Steps to reproduce
```bash
export DONSETCH_CACHE_DIR=/tmp/ds-cache-test
donsetch keys add tinyfish <token>
donsetch keys default tinyfish
donsetch search "cache test 42" --max-results 3 --json | jq -c '{provider:.meta.provider, cached:.meta.cached}'
donsetch search "cache test 42" --max-results 3 --json | jq -c '{provider:.meta.provider, cached:.meta.cached}'
```

Observed: `{"provider":"tinyfish","cached":false}` **twice**.

Control (same profile, no BYOK key, `default = local`): first call `cached:false`, **second call `cached:true`** — the cache works, it is simply skipped on the BYOK path.

## Expected
BYOK results cached with the same semantics/TTL as local results (30 min web / 5 min news / 15 min code), **or** a documented note that BYOK calls bypass the cache.

If this is intentional (e.g. to respect provider freshness or a BYOK contract), feel free to close this — a one-line note in the README would be enough for us.

## Impact
Agents repeat queries (ours do — we had to add a de-duplication wrapper). With a metered provider each repeat is a silent re-bill; with a free tier it burns quota for no benefit.

## Code pointers (3.6.7)
- BYOK search returns without consulting or persisting the cache: `src/search/byok/mod.rs::ByokSearcher::search` (the outcome carries `cached: false`), and the BYOK branch of `src/mcp/server/search_tool.rs`.
- Local cache read/write: `src/search/mod.rs` (the functions the local path already uses).

mnaza:
Closes #195.

## The bug

The local search path reads and writes a TTL'd result cache and reports `meta.cached = true` on a repeat. The BYOK path did neither — `search_outcome` called `daemon.byok.search(...)` and returned without ever consulting or populating the cache. So `meta.cached` was `false` on every call, and a **metered provider was re-billed for every identical query** inside the freshness window (30 min web / 5 min news / 15 min code).

Reproduction from the issue:

```
$ donsetch search "cache test 42" --json | jq -c '{provider:.meta.provider, cached:.meta.cached}'
{"provider":"tinyfish","cached":false}
$ donsetch search "cache test 42" --json | jq -c '{provider:.meta.provider, cached:.meta.cached}'
{"provider":"tinyfish","cached":false}   # ← should be cached:true, no provider call
```

## The fix

BYOK results now ride the **same** cache, TTL and disk file as local results, in their own key namespace:

- `Searcher::byok_cache_get` / `byok_cache_put` key entries as `byok|<norm_query>|<intent>` (local uses `<norm_query>|<intent>`). The extra segment isolates the two paths — flipping `default` between local and a provider can never cross-serve. `cache_ttl(intent, query)` is applied identically; on a hit the outcome carries `cached = true` and the provider recovered from the stored report, which `render_meta` already surfaces as `meta.cached` / `meta.provider`.
- `byok_search_cached` wraps **both** BYOK entry points (BYOK-first and the local-first fallback) so caching can't drift between them: read the cache first (a hit returns without calling the provider), else call the provider, store the full top slice **before** `site:` filtering, then filter and prewarm exactly as the old code did.

No change to the cache tuple or the on-disk format — byok entries are the same shape and round-trip through `load_cache_disk` unchanged (its `rsplit_once('|')` still parses the intent and never lands a byok row in a local slot).

Now the second call returns `{"provider":"tinyfish","cached":true}` with no provider request.

## Verification

New unit tests pin the roundtrip (`cached=true`, provider preserved, `weak=false`) and namespace isolation (a byok write never populates the local slot; a different intent misses). Full `cargo nextest run` **1107/1107** (1 skipped); `clippy --all-targets` and `fmt --check` clean.

The tests are hermetic: the search cache persists via `paths::cache_dir()` (honors `DONSETCH_CACHE_DIR`), so the test harness points it at a throwaway dir and clears the in-memory map after construction — otherwise a `byok_cache_put` would pollute the real user cache and a later process-isolated run would load it.

## Known limits, left out by scope (all strict improvements over caching nothing)

- Two identical BYOK queries **in flight at once** still both bill — the local path has single-flight admission, BYOK does not. Sequential repeats (the issue) are fully covered; concurrent single-flight would be a separate enhancement.
- A provider result flagged `degraded` is cached for the TTL like any other (the local path has an `ok_engines>=2 && total>=8` poisoning guard that has no BYOK equivalent).
- An entry stored at `max=3` re-served at `max=10` returns the 3 the provider was billed for (local always merges 12; BYOK can't without billing for more).

Happy to also just document a cache-bypass instead if that matches the BYOK contract you intend — but caching with the local TTL semantics seemed to be the ask.

### DIFF
--- a/src/mcp/server/search_tool.rs
+++ b/src/mcp/server/search_tool.rs
@@ -399,20 +399,8 @@ pub(super) async fn search_outcome(
 
     // BYOK-first mode: try providers, fall back to local.
     if byok_configured && !local_first {
-        match daemon.byok.search(query, max, intent).await {
-            Ok(out) => {
-                let mut out = out;
-                // Issue #190: site: queries dawn on BYOK results too,
-                // and prewarm only the rows that survive the filter.
-                crate::search::site_filter(query, &mut out.results);
-                // v4 phase 2.1: the warm handoff is not a local-search
-                // privilege. Provider results carry their own titles
-                // and snippets, so the reply does not wait: prewarm
-                // fires detached and parks bodies while the model
-                // reads the results (law 5: zero added latency).
-                daemon.searcher.spawn_prewarm(&out.results);
-                return Ok(out);
-            }
+        match byok_search_cached(daemon, query, max, intent).await {
+            Ok(out) => return Ok(out),
             Err(e) => {
                 if std::env::var_os("DONSEEK_DEBUG").is_some() {
                     eprintln!("[byok] all providers exhausted, falling back to local: {e}");
@@ -432,17 +420,8 @@ pub(super) async fn search_outcome(
                 if std::env::var_os("DONSEEK_DEBUG").is_some() {
                     eprintln!("[byok] local search failed, trying BYOK fallback: {e}");
                 }
-                match daemon.byok.search(query, max, intent).await {
-                    Ok(out) => {
-                        // Issue #190: the site: scan has to reach the
-                        // fallback path too, or the local-first mode
-                        // leaks rows off-domain.
-                        let mut out = out;
-                        crate::search::site_filter(query, &mut out.results);
-                        // Same detached prewarm as the BYOK-first path.
-                        daemon.searcher.spawn_prewarm(&out.results);
-                        Ok(out)
-                    }
+                match byok_search_cached(daemon, query, max, intent).await {
+                    Ok(out) => Ok(out),
                     Err(e2) => Err(SearchFailure {
                         cause: format!("local ({e}); byok ({e2})"),
                         byok_tried: true,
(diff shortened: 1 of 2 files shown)


dondai44423:
will be on this later.

currently removing some bloat that i introduced with v4 changes (web answer and web memory) are pretty shit and useless feature from my testing

PR #201, opened 2026-09-11
build: Windows cross-check from Linux, `just win-check`

Mart-Bogdan:
## Summary

#196 broke Windows twice in one day, both times in `cfg`-gated code that never compiles on a Linux box. This adds `just win-check`: clippy for `x86_64-pc-windows-gnu` run from Linux, no linkage, at both ends of the feature matrix (`win-check-core` with `--no-default-features`, `win-check-full` with `ocr,rerank,http`), `--all-targets`, `-Dwarnings`. Both halves fail on the two commits that broke Windows today. Not part of `just all`, since it needs a mingw cross toolchain on top of the normal prerequisites; CONTRIBUTING.md has the install lines for Debian/Ubuntu and the distro-neutral equivalents. Three environment workarounds make the cross dependency graph buildable (nasm `-pthread` leak from CMake 3.28, mingw-GCC include dir for bindgen, no ONNX prebuilts for windows-gnu); each is explained inline in the Justfile. Type-checking is the ceiling: a cross `cargo build` fails to link in BoringSSL even featureless, so Windows CI proper stays authoritative.

## Type

- [ ] `feat` — new feature
- [ ] `fix` — bug fix
- [x] `chore` — deps, CI, tooling
- [ ] `test` — tests only

## Checks

No Rust source changes: Justfile, CONTRIBUTING.md, CHANGELOG.md and one POSIX shell script. The three local checks are unaffected by this PR; CI ticked once it reports.

## Breaking changes

None. New recipes only; `just all`, `just lint` and CI are untouched.

## Test plan

On WSL2 Ubuntu 24.04 (mingw-w64 13, cmake 3.28.3, nasm 2.16.01, rustc 1.98.1):

- `just win-check` green on the branch head, both halves. First run compiles BoringSSL for the mingw target (a few minutes), warm runs take seconds.
- Regression check against the two real breaks. Checked out each commit detached, ran both halves with the same env and flags (wrapper copied to `/tmp`, since those commits predate it):
  - `55bda0b`: both halves fail with `error[E0433]: cannot find type PathBuf` at `src/profile/mod.rs:382`, the Windows-only path that `0462816` fixed.
  - `f3b6005`: both halves fail with `unused import: spawn_probe_with_timeout` at `src/profile/mod.rs:727` in the lib test target, the break that `dfbf38c` fixed. `--all-targets` and `-Dwarnings` are load-bearing there: the code is in a test, and without `-Dwarnings` an unused import is only a warning.
- One nuance for anyone comparing against the Windows CI log: rustc reports the first error it hits, so the message can differ for the same line. `55bda0b` also had a type mismatch on that line, but name resolution fails first, so the cross-check shows E0433. Same break, different first symptom.
- Missing-toolchain path: with `x86_64-w64-mingw32-gcc` absent, `just win-check-core` prints the install hint and exits 1 before cargo runs.
- Not verified on macOS. The recipe is POSIX shell and the workarounds are toolchain-level, so it should work with a Homebrew `mingw-w64`, but nobody has run it there.

Not in this PR: a Dockerfile wrapping `win-check`, for anyone who would rather not install mingw locally. Happy to add one if wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ V4 work in progress on `master`. Nothing below ships through a release
 channel until the v4.0.0 release train.
 
 ### Added
+- `just win-check`: type-checks the crate for `x86_64-pc-windows-gnu`
+  from Linux (clippy, no linkage), both `--no-default-features` and
+  the full feature set, so `#[cfg(windows)]` breakage from a
+  Linux-only change surfaces before the push instead of in Windows
+  CI. Needs mingw-w64; see CONTRIBUTING.md.
 - `web_screenshot` MCP tool: a rendered PNG of a page through the
   existing tier-2 browser (url, full_page, wait_ms). The capture is
   in-process only; the MCP result carries an image content block and
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,51 @@ All three must pass before a PR can merge. CI runs the same checks on Linux, mac
 
 The same tasks are wrapped as [`just`](https://just.systems) recipes — `just test`, `just lint`, `just fmt-check`, `just smoke`.
 
+### Verifying Windows compilation from Linux
+
+Linux-only changes regularly broke Windows CI in `cfg`-gated code that never
+compiled locally. `just win-check` type-checks the whole crate for
+`x86_64-pc-windows-gnu` without linking — both ends of the feature matrix
+(full `ocr,rerank,http` and `--no-default-features`), all targets,
+`-Dwarnings` — so those breaks surface before the push. It is not part of
+`just all` because it needs a cross toolchain on top of the normal build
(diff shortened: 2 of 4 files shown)


dondai44423:
Merged. I also ran the recipes live on this machine instead of trusting the docs (gcc 14.2 mingw, clang, rust-toolchain pinned), and it caught two real gaps which I ported in a follow-up (9746150):

- gcc 14's private stddef carries its own max_align_t and clashes with clang's copy inside bindgen, so both recipe halves now define __CLANG_MAX_ALIGN_T_DEFINED. Harmless on gcc 12/13, required on 14+.
- The rustup target guard read the default toolchain while this repo's rust-toolchain.toml pins 1.98, so a correct rustup target add could false-fail. The guard now resolves the active toolchain first.

Both halves verified green here after the ports. Nice PR, the nasm wrapper is the clean fix for that CMake 3.28 leak.

PR #202, opened 2026-09-12
fix(supervisor): keep the replayed request across a second silent death; bound the live history

mnaza:
Follow-up to 5a0675b, which started replaying a child's unacked stdin history so a request buffered into a child that dies silently (the write lands in the pipe, no EPIPE fires, the death only surfaces on the idle poll) still reaches the replacement. Two gaps remained.

## 1. The replay survived exactly one death

On respawn the held `pending` is written to the new child, and then `written` was `clear()`ed:

```rust
if !pending.is_empty() && stdin.write_all(&pending).is_ok() {
    let _ = stdin.flush();
    pending.clear();
}
written.clear();   // ← the just-replayed bytes are forgotten
```

If the replacement **also** buffers-then-dies before consuming that request — a crash loop where each child dies during startup after its pipe fills — the next idle poll checks `written.is_empty()`, finds it empty, and drops the request one restart deeper. The exact failure 5a0675b set out to prevent, just one death later.

The replayed bytes are the new child's unacked history too, so seed `written` with them instead of clearing:

```rust
if !pending.is_empty() && stdin.write_all(&pending).is_ok() {
    let _ = stdin.flush();
    written = std::mem::take(&mut pending);   // survive a second death
} else {
    written.clear();
}
```

## 2. The "bounded to 1 MiB" claim didn't hold for the live buffer

The commit message and comment say the history is bounded to 1 MiB, but only the replay **output** was bounded — `replay_window` caps the `Vec` it produces at a death. The live accumulator was uncapped:

```rust
written.extend_from_slice(&bytes);   // grows forever for a healthy child
```

A healthy long-lived supervised daemon consumes every request immediately, but the supervisor still appended every stdin byte the client ever sent to `written` and never freed it. `record_written` now caps the history to `REPLAY_WINDOW` on every append (keeping the tail — the newest requests are the ones a replacement needs), so the documented bound is true of the live buffer, not just the death-time output. `replay_window` delegates to it.

## Verification

- `buffered_request_survives_two_consecutive_silent_deaths`: children 1 and 2 accept the write and die at 300 ms, child 3 serves. **RED** on the current `written.clear()` (the sink comes back empty — the request is lost on the second death); GREEN with the seed.
- `record_written_bounds_the_history_to_the_replay_window`: deterministic — write well past the window in chunks, assert the history stays `<= REPLAY_WINDOW` and keeps the tail.

Full `cargo nextest run` **1116/1116** (1 skipped); `clippy --all-targets` and `fmt --check` clean. The new duplicate-delivery on a seeded replay is the same tradeoff the commit already chose ("duplicate delivery preferred over a lost request"), now bounded.

### DIFF
--- a/src/mcp/supervisor.rs
+++ b/src/mcp/supervisor.rs
@@ -152,9 +152,17 @@ where
             // (Write failure: this child already died; keep pending.)
             if !pending.is_empty() && stdin.write_all(&pending).is_ok() {
                 let _ = stdin.flush();
-                pending.clear();
+                // The replayed bytes are THIS child's unacked history
+                // too: if it also dies before consuming them (a crash
+                // loop where each child buffers-then-dies), the idle
+                // poll must replay them again, not find an empty
+                // history and drop the request one restart deeper.
+                // Seed the history with what we just wrote instead of
+                // clearing it.
+                written = std::mem::take(&mut pending);
+            } else {
+                written.clear();
             }
-            written.clear();
             child = Some((c, stdin, Instant::now()));
         }
 
@@ -163,7 +171,7 @@ where
         match rx.recv_timeout(POLL) {
             Ok(In::Data(bytes)) => {
                 if stdin.write_all(&bytes).is_ok() {
-                    written.extend_from_slice(&bytes);
+                    record_written(&mut written, &bytes);
                     let _ = stdin.flush();
                 } else {
                     // Child died under this write : hold this span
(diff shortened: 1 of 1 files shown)


dondai44423:
Verified and merged. Both gaps are real against the tree:

- On respawn the replay wrote the held bytes into the new child and cleared the history, so a replacement that also buffered-then-dies dropped the request one restart deeper. Reproduced it myself: the old semantics fail your two-death test, the seed passes it.
- The live accumulator grew without limit on a healthy child; the 1 MiB claim now holds on every append, not just at death time.

Gate x2 green on your branch (1155 + 1 skip) before the merge, changelog noted. Nice follow-up, exactly the crash loop the first fix left one death deep.

PR #204, opened 2026-09-12
fix(config): redact secrets from diagnostics

Mart-Bogdan:
Looks like PR #203 was auto-closed as soon as as #196 was merged, as this destoyed branch `config-consolidation`, on which #203 was pointing to. 

Can't reopen that PR as branch is destroyed:
image

So I am opening new one. I've already rebased it.

FYI: @adaaaaaaaaaaaaaaaaaaaaaa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ V4 work in progress on `master`. Nothing below ships through a release
 channel until the v4.0.0 release train.
 
 ### Changed
+- `donsetch config show` redacts bearer tokens and credential-bearing proxy
+  values, including proxy pools, while still showing whether each field is
+  configured and where its value came from.
 - Every runtime env read now flows through the typed config. The old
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and merged. The transport token was a real gap in my d833e47 redaction (config show printed it verbatim; the bot's high was half-closed until this). Also verified your other two by running them: the redaction test fails against the old rfind-less mask (the password-with-at case) and the token arm, and passes with your fix on the merged master. Gate x2 green (1165 + 1 skip). Merged as a squash.

dondai44423:
Verified and merged (d61f5d3). The transport token was a real gap in my d833e47 redaction: config show printed it verbatim, so the bot's high was only half-closed until this. Also verified by running: the upgraded test fails against the old find-based mask (the password-with-at case) and the token arm, passes with your fix on the merged master. Gate x2 green (1165 + 1 skip).

PR #206, opened 2026-09-12
fix(profile): bound browser version probe lifecycle

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

Completes the browser version-probe timeout started in `cf83850` during #196. A probe whose parent exited while a descendant retained stdout could leave the reader join blocked past the advertised deadline; continuous output was also unbounded.

- apply one post-spawn deadline to both child completion and stdout EOF
- replace the blocking reader-thread join with bounded, non-blocking capture
- cap probe output at 64 KiB
- contain descendants in a Unix process group or the existing Windows kill-on-close Job Object abstraction
- assign the Windows process while suspended, before descendants can escape the Job Object
- cover both event orders: a live parent and a parent that exits while a descendant retains stdout

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo nextest run --cargo-profile ci --features ocr,rerank,http` passes: 1158 passed, 1 skipped
- [x] `cargo clippy --all-targets --features ocr,rerank,http -- -Dwarnings` passes

## Breaking changes

None. Successful version parsing and the three-second public timeout are unchanged.

## Test plan

- parent remains alive with stdout open, or closes stdout before exiting
- parent exits while a descendant retains stdout
- valid output, invalid output, invalid UTF-8, and non-zero exit
- exact 64 KiB boundary, 64 KiB + 1, and an infinite output stream
- successful parent with a detached descendant verifies process-tree cleanup
- existing public-wrapper descendant timeout regression
- Windows Job Object fixtures compile for MSVC; runtime execution remains Windows CI-authoritative

The capture loop now waits for process exit and pipe EOF under the same deadline. Teardown closes the process-tree container and reaps the direct child. Windows Job Object creation and teardown are shared with DonGhost rather than maintained as a second implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ channel until the v4.0.0 release train.
 - `donsetch config show` redacts bearer tokens and credential-bearing proxy
   values, including proxy pools, while still showing whether each field is
   configured and where its value came from.
+- Browser version probes now apply one deadline to process exit and bounded
+  stdout capture, terminate inherited-pipe descendants on Unix and Windows,
+  and cap captured output at 64 KiB.
 - Every runtime env read now flows through the typed config. The old
   env names keep their exact historical trigger semantics and still
   work, but they are deprecated and will be removed at the v4
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ rerank = ["dep:ort", "dep:tokenizers", "dep:ndarray", "dep:num_cpus", "reqwest/b
 http = ["dep:axum", "dep:tower-http"]
 
 [target.'cfg(windows)'.dependencies]
-windows-sys = { version = "0.61", features = ["Win32_System_JobObjects", "Win32_System_Threading", "Win32_Security", "Win32_Foundation", "Win32_System_Console", "Win32_System_Registry"] }
+windows-sys = { version = "0.61", features = ["Win32_System_JobObjects", "Win32_System_Threading", "Win32_Security", "Win32_Foundation", "Win32_System_Console", "Win32_System_Registry", "Win32_System_Pipes"] }
 
 [dev-dependencies]
 rcgen = "0.14"
--- a/src/ghost/proc.rs
+++ b/src/ghost/proc.rs
@@ -42,7 +42,7 @@ pub struct Proc {
     #[cfg(windows)]
     proc_handle: fnd::HANDLE,
     #[cfg(windows)]
-    job: fnd::HANDLE,
+    job: KillOnCloseJob,
 }
 
 // ntdll process suspend/resume : always loaded, linked directly.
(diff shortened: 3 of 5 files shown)


Mart-Bogdan:
**`job_capture_accepts_finite_output_and_bounds_floods` fails deterministically on Windows — the `"detached"` case, not resume**

Reproduced locally on `0d06429` (Windows 11, toolchain 1.98, nextest, featureless build to match the Windows CI lane), and it is the same panic the first CI run showed:

```
thread 'profile::version_probe::windows_tests::job_capture_accepts_finite_output_and_bounds_floods' panicked at src\profile\version_probe.rs:430:63:
called `Result::unwrap()` on an `Err` value: "browser version probe timed out before process exit and stdout EOF"
```

The `ResumeThread` change in the latest push doesn't affect this. `"valid"` and `"flood"` pass before it, so the process was already resuming fine; the failure is that the `"detached"` descendant still holds the probe's stdout pipe.

Cause: `fixture("descendant").stdout(Stdio::null())` only redirects `STD_OUTPUT_HANDLE`. Rust's `Command::spawn` on Windows calls `CreateProcessW` with `bInheritHandles = TRUE` and no `PROC_THREAD_ATTRIBUTE_HANDLE_LIST`, so every inheritable handle in the fixture process is copied into the descendant's handle table anyway, including the pipe write-end the fixture received as *its* stdout. The descendant holds that handle for 30 s, `capture` never sees EOF, and the probe times out. This is std behaviour, not a bug in the probe, but it means the test's premise ("a detached descendant does not hold the pipe") is false on Windows.

Two ways out, and they're really a design choice:

1. **Keep fail-closed, fix the fixture.** Spawn the descendant with a `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` that excludes the inherited stdout (via `CommandExt::raw_attribute`), or `SetHandleInformation(HANDLE_FLAG_INHERIT, 0)` on the fixture's own stdout handle before spawning. The test then proves what the name says.
2. **Make the capture lenient.** Once `try_wait` reports the parent exited 0, drain whatever is readable and parse it, then let the Job Object / process-group teardown kill the leftovers. That is closer to what a version probe actually needs, and it also covers any real browser whose crash handler or helper inherits stdout. It would flip `parent_exit_does_not_cancel_stdout_deadline` from "expect timeout" to "expect 151.0.0.0".

I'd lean towards 2, but either is fine as long as the Windows test matches the chosen semantics.

Separately, I ran the probe against installed `chrome.exe` and `msedge.exe` on this machine with both master and this branch: neither answers `--version --headless=new` on Windows (no stdout, still running after 5 s), so `doctor` lands on the registry fallback in both cases. Not a regression from this PR, just noting the Windows probe path is effectively untested by real binaries.

dondai44423:
Verified and merged. I read the whole diff against master and ran your five unix tests on the merged tree (all green, the parent-exit one in 0.27s): the real bug you completed is the stdout join after a successful child exit having no deadline at all, so a descendant holding the pipe hung the probe for the descendant's lifetime. The single-deadline capture with the process group / suspended Job Object assignment covers both event orders, the 64 KiB cap is bounded on both sides, and the KillOnCloseJob consolidation keeps ghost teardown behavior identical. Gate x2 green on the merged tree (1169 + 1 skip).

PR #207, opened 2026-09-12
fix(http): fail closed on ambiguous bearer tokens

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- Reject ambiguous modern `transport.token` values from TOML and `DONSETCH_TRANSPORT__TOKEN`; whitespace-only and non-UTF-8 values now fail configuration loading instead of silently disabling authentication.
- Preserve the legacy `DONSETCH_HTTP_TOKEN` contract: exact non-empty UTF-8 values remain authentication-on and are compared without trimming, while an exact empty value remains authentication-off.
- Validate the auth/CORS combination before daemon construction starts background work or touches persistent state.
- Keep the token contract centralized and ensure diagnostics never echo the configured secret.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo nextest run --cargo-profile ci --features ocr,rerank,http` passes (1,156 passed, 1 skipped)
- [x] `cargo clippy --all-targets --features ocr,rerank,http -- -Dwarnings` passes (zero warnings)

## Breaking changes

None for valid configurations. Modern token values containing whitespace, non-visible bytes, or non-UTF-8 data now fail explicitly instead of being normalized or ignored.

## Test plan

- TOML and modern-env tokens containing whitespace are rejected without appearing in the error text.
- A whitespace-only modern token cannot collapse to authentication-off.
- Non-UTF-8 modern env values fail closed on Unix.
- Visible-ASCII modern tokens load unchanged.
- Legacy padded UTF-8 tokens retain exact-match behavior; a higher-precedence explicit modern empty token still disables auth.
- Legacy non-UTF-8 values retain their historical ignored fallback and emit only a static warning.
- HTTP bearer matching verifies the exact legacy value rather than a trimmed form.
- The complete CI-profile Nextest suite, Clippy with HTTP enabled, formatting, and diff checks pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,6 +170,11 @@ channel until the v4.0.0 release train.
 
 ### Fixed
 
+- HTTP bearer configuration now fails closed: modern TOML/env tokens reject
+  whitespace and non-visible bytes without echoing the secret, while legacy
+  text tokens retain their exact historical value instead of being trimmed
+  into disabled authentication. CORS/auth validation also runs before daemon
+  work.
 - The supervisor's replay now survives a crash loop: the bytes
   replayed into a replacement were cleared from the unacked
   history, so a second silent death dropped them (reported by
--- a/src/config.rs
+++ b/src/config.rs
@@ -324,12 +324,19 @@ pub fn load() -> Result<Loaded, ConfigError> {
     // Validate the file layer on its own so unknown keys and bad types
     // fail loudly with the file path attached, exactly once.
     if let Some(file) = &file_layer {
-        deserialize_layer_table(file).map_err(|m| ConfigError::File {
-            path: file_path
-                .as_ref()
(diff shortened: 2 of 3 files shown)


dondai44423:
Verified and merged (30b8b34). The fail-closed contract is the right shape: a whitespace or non-UTF-8 transport token used to resolve to an ambiguous trim-or-off state at request time; now the config layer rejects it loudly and the error never echoes the value, while the legacy DONSETCH_HTTP_TOKEN path keeps its exact untrimmed compare (your no-trim test pins that). I ran the discriminator myself: shimming the validator out fails your strict test in the loader, and the union tree gate x2 is green (1173 + 1 skip). Also the auth validation running before daemon construction closes the ordering hole I had in run().

PR #208, opened 2026-09-12
fix(config): preserve browser override presence

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- preserve the winning config origin in the installed singleton so legacy browser overrides retain presence semantics
- distinguish absent overrides, legacy empty values, and modern empty resets
- keep Chromium, CloakBrowser, and Playwright paths literal
- document the behavior under Unreleased

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo nextest run --cargo-profile ci --features ocr,rerank` (1,149 passed, 1 skipped)
- [x] `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings`

## Breaking changes

None. Existing legacy presence semantics are restored; modern empty values remain the explicit reset form.

## Test plan

- cover absent browser overrides falling back to discovery or ambient values
- cover legacy empty values remaining explicitly supplied
- cover modern empty values clearing lower-precedence legacy overrides
- cover an explicit empty Playwright value suppressing the ambient mirror

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,10 @@ channel until the v4.0.0 release train.
   text tokens retain their exact historical value instead of being trimmed
   into disabled authentication. CORS/auth validation also runs before daemon
   work.
+- Browser path overrides now preserve source presence across the typed config:
+  legacy empty values keep their historical override semantics, while modern
+  empty values reset to discovery or ambient defaults. Explicit Chromium,
+  CloakBrowser and Playwright paths remain literal.
 - The supervisor's replay now survives a crash loop: the bytes
   replayed into a replacement were cleared from the unacked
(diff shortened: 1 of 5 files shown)


dondai44423:
Verified and landed. The presence semantics are the real gap the typed config left behind: DONGHOST_CHROME set-but-empty silently degraded to system discovery, a modern empty could not reset a lower legacy override, and the playwright mirror revived the ambient var after an explicit empty reset. The origin-carrying OnceLock<Loaded> is the right way to keep source presence without re-reading env in runtime consumers; programmatic install() keeps value-only semantics. Your test, plus the scoped config batch, ran green on the union tree; full gate x2 green (1200 + 1 skip).

PR #209, opened 2026-09-12
fix(config): preserve legacy control composition

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- preserve the historical `DONSETCH_NO_ROUTE_MEMORY` precedence over `DONSETCH_ROUTE_MEMORY_READONLY`
- preserve the historical bypass-cache ordering where a valid positive legacy TTL re-enables the cache after `DONSETCH_BYPASS_CACHE=0`
- treat legacy TTL zero as cache-disabled while keeping absent and invalid values distinct
- retain TOML and modern environment precedence
- document the restored composition under Unreleased

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] full `ocr,rerank,http` Nextest suite (1,161 passed, 1 skipped)

## Breaking changes

None. This restores pre-migration legacy behavior.

## Test plan

- table-test route-memory kill-switch/read-only ordering
- table-test bypass cache enabled, disabled, positive TTL, zero TTL, invalid TTL, and absent TTL
- assert effective values, winning origins, and warning behavior
- verify historical semantics against commits `04663e6` and `b709588`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,13 @@ channel until the v4.0.0 release train.
   text tokens retain their exact historical value instead of being trimmed
   into disabled authentication. CORS/auth validation also runs before daemon
   work.
+
+- Legacy route-memory and bypass-cache controls compose in their original
+  order under the typed config: the route-memory kill switch beats read-only,
+  while a valid legacy bypass TTL applied after `DONSETCH_BYPASS_CACHE=0`
+  re-enables the cache (and TTL zero disables it). TOML and modern env
+  overrides retain their higher precedence.
+
 - The supervisor's replay now survives a crash loop: the bytes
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and landed. Both compositions were real legacy regressions the typed config introduced: NO_ROUTE_MEMORY used to lose to READONLY on the overwrite order (kill switch must win, same rule the pool uses), and a valid legacy TTL after BYPASS_CACHE=0 historically re-enabled the cache while TTL=0 disabled it; out-of-range stays warning-only. Proven discriminating on my side: shimming the precedence back to independent ifs fails your composition test, and both tests run green on the union tree; full gate x2 green (1200 + 1 skip).

PR #210, opened 2026-09-12
fix(search): consume typed reranker thread config

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

Consume the validated `search.rerank_threads` integer directly when initializing the ONNX reranker. This removes an initialization-only string leak and reparse, preserves `0 = auto`, reports configured values as coming from layered config, and documents the fix under Unreleased.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] full test and lint matrix passed in CI for the code commit
- [x] `cargo test --profile ci --features rerank --lib rerank_threads` (3 passed)
- [x] `cargo clippy --profile ci --lib --features rerank -- -Dwarnings`

## Breaking changes

None.

## Test plan

- verify automatic selection clamps constrained processes
- verify unconstrained hosts preserve the ONNX default
- verify configured values `1`, `2`, and `64`
- verify the log reports `source: layered config`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,6 +176,9 @@ channel until the v4.0.0 release train.
 
 ### Fixed
 
+- The ONNX reranker now consumes the validated `search.rerank_threads` value
+  directly instead of leaking and reparsing a string, and reports configured
+  values as coming from the layered config. `0 = auto` behavior is unchanged.
 - HTTP bearer configuration now fails closed: modern TOML/env tokens reject
   whitespace and non-visible bytes without echoing the secret, while legacy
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and landed. The Box::leak string round-trip for an already-validated u32 was pure waste and the parse machinery duplicated what config load already enforces; 0 = auto survives unchanged. The layered-config source label in the log line is also more accurate now that the knob takes modern env names. Scoped rerank tests green on the union tree; full gate x2 green.

PR #211, opened 2026-09-12
fix(config): normalize browser backend aliases

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- normalize browser backend casing and surrounding whitespace at the typed serde boundary
- keep `BrowserBackend` aliases as the single alias registry
- preserve unknown legacy values so an effective invalid selection fails closed
- allow valid higher-precedence modern configuration to override an invalid legacy value
- remove the obsolete duplicate parser from the browser resolver
- avoid echoing rejected backend values and document the change under Unreleased

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] full Nextest suite (1,161 passed, 1 skipped)

## Breaking changes

None. Valid aliases are unchanged; effective invalid legacy values now fail during config loading.

## Test plan

- cover mixed-case and whitespace-normalized aliases
- cover invalid effective legacy values failing without echoing the value
- cover a valid modern override winning over an invalid legacy value
- preserve the historical legacy empty-string `auto` behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,6 +176,10 @@ channel until the v4.0.0 release train.
 
 ### Fixed
 
+- Browser backend aliases now share the typed enum as their single registry
+  and normalize casing and surrounding whitespace across config sources.
+  Unknown effective legacy values fail closed instead of silently becoming
+  `auto`; a valid higher-precedence modern setting still overrides them.
 - HTTP bearer configuration now fails closed: modern TOML/env tokens reject
   whitespace and non-visible bytes without echoing the secret, while legacy
   text tokens retain their exact historical value instead of being trimmed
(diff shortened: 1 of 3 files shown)


dondai44423:
Verified and landed. The single-registry consolidation is the right shape: the typed enum owns every alias, the legacy mapper's silent warn-and-fallback is gone (an unknown backend silently becoming auto is exactly the wrong browser being launched with no signal), and a valid modern override still beats an unknown legacy value instead of poisoning it. The value-free error and the normalization table are both pinned by tests. Your scoped tests plus my shim discriminator (unknown value forced back to auto = your fail-closed test red) confirm it; full gate x2 green.

PR #212, opened 2026-09-12
fix(config): validate TOML before overrides

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

Validate the deserialized TOML source before applying higher-precedence environment overrides. Invalid strict file values can no longer be hidden by valid environment overrides. The fix reuses the existing validation source of truth, preserves valid precedence and provenance, keeps legacy values permissive, and is documented under Unreleased.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] targeted config validation tests

## Breaking changes

None. Modern TOML and environment sources remain strict; legacy values remain warning-only when invalid.

## Test plan

- prove an invalid TOML port is rejected even when a valid modern environment value would override it
- prove the error retains file-path attribution
- prove a valid TOML value is still overridden with environment provenance
- verify invalid legacy values remain warning-only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,10 @@ channel until the v4.0.0 release train.
   text tokens retain their exact historical value instead of being trimmed
   into disabled authentication. CORS/auth validation also runs before daemon
   work.
+- Typed config validation now rejects an out-of-policy TOML value with
+  file attribution even when a valid higher-precedence env value would
+  otherwise hide it; legacy out-of-range values remain warning-only.
+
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and landed. File-layer validation running before the env override can mask it closes a real loudness hole: an out-of-policy TOML value loads fine today and goes live the moment the overriding var is unset, with zero signal. Attribution stays on the file, the valid-override path keeps its env origin. Shim-verified on my side (dropping the validate call fails your test), scoped batch + full gate x2 green.

PR #213, opened 2026-09-12
fix(config): reject ambiguous modern env input

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- reject non-UTF-8 values for recognized modern environment variables
- reject distinct environment names that normalize to the same config key
- parse an immutable environment snapshot so collision handling is independent of OS iteration order
- keep diagnostics value-free and preserve legacy semantics
- compose with the HTTP token validation from #207
- document the behavior under Unreleased

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] 24 config tests
- [x] full suite (1,165 passed, 1 skipped)
- [x] full-feature all-target Clippy with `-Dwarnings`

## Breaking changes

None for valid configuration. Previously ambiguous modern environment input now fails loudly instead of depending on iteration order or being silently ignored.

## Test plan

- cover recognized non-UTF-8 modern values without echoing bytes
- cover normalized-name collisions independently of snapshot order
- cover value-free type errors
- verify legacy environment semantics remain unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,12 @@ channel until the v4.0.0 release train.
   text tokens retain their exact historical value instead of being trimmed
   into disabled authentication. CORS/auth validation also runs before daemon
   work.
+- Modern `DONSETCH_<SECTION>__<KEY>` variables now fail loudly when a
+  recognized value is not UTF-8 or when distinct names normalize to the same
+  config key. Diagnostics identify the variable names without exposing their
+  values, and collision handling no longer depends on environment iteration
+  order.
+
 - The supervisor's replay now survives a crash loop: the bytes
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and landed. Three honest wins: a recognized modern var whose value is not UTF-8 now fails instead of silently skipping (the config would have run on defaults while the operator believes their knob applied), distinct names normalizing to one key now error deterministically instead of whoever wins the OS iteration order, and type errors name the variable without echoing the value. The immutable-snapshot injection also makes the layer order-independent in tests, which is how I verified the collision error is stable both ways. Shim-verified (collision detection dropped = your test red), scoped batch + full gate x2 green.

PR #214, opened 2026-09-12
fix(config): render markdown tables per section

adaaaaaaaaaaaaaaaaaaaaaa:
## Summary

- render `config show --markdown` as one complete Markdown table per config section
- keep the fieldbook as the single source of truth and preserve one row per knob
- replace the pipe-count-only regression test with a structural section/table test
- document the fix under Unreleased

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo nextest run --cargo-profile ci --features ocr,rerank,http` (1,162 passed, 1 skipped)

## Breaking changes

None. The Markdown representation changes from a broken global table to valid per-section tables; knob order and values are unchanged.

## Test plan

- structurally require every section label to be followed by a contiguous header and delimiter
- require exactly one ordered row for every fieldbook knob
- keep embedded pipes escaped
- render with Pandoc/GFM and verify 13 tables with 71 data rows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,9 @@ channel until the v4.0.0 release train.
   text tokens retain their exact historical value instead of being trimmed
   into disabled authentication. CORS/auth validation also runs before daemon
   work.
+- `donsetch config show --markdown` now emits one complete Markdown table per
+  config section, so section labels no longer turn the following knob rows
+  into plain paragraphs.
 - The supervisor's replay now survives a crash loop: the bytes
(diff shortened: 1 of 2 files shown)


dondai44423:
Verified and landed. The section label mid-table broke every markdown renderer's table context, so rows after the first label rendered as plain text. One complete table per section with the structural test replacing the pipe-count heuristic is the right shape, and the escaped-pipe case stays covered. Shim-verified (reverting to the single header fails your structural test), full gate x2 green.

PR #215, opened 2026-09-12
Feat/transport dead resurrection

imonlinux:
Summary

Wayback resurrection today serves dead pages and hard walls, but refuses the archetypal dead link: the site whose DNS is gone, whose port refuses, whose TLS handshake hits a parked domain's certificate. This PR extends the resurrection gate to those transport-dead classes and, in exchange, makes every give-up on the archive path honest about which stage failed.

**The policy change, argued explicitly:** `fetch_single` today treats transport errors as "maybe dead" and never resurrects them. That is correct for ambiguous transients, and this PR keeps every one of those excluded: timeouts, resets and protocol errors do not resurrect, because a snapshot must never launder an unknown into fake certainty, and a reset can be an IP-level block that an archived copy would paper over. But `tls` (no handshake), `dns` (name gone) and `refused` (port closed) are not ambiguous. They are the three ways a domain actually dies, and today they return a bare error while the archive still holds the content. The gate reads a new transport class recorded on every status-0 error (`structuredContent.fetch_error`), so the boundary is explicit and testable: exactly `tls | dns | refused` resurrect, everything else does not.

Three honesty fixes ride along (same subsystem, so one PR):

1. **"Never archived" was claimed from a lossy, scheme-strict source.** The availability API alone no longer decides: an empty answer falls through to the complete CDX index (scheme-canonical, `filter=statuscode:200`, last 5 captures), and the snapshot URL is rebuilt from the capture's `original` column, because rebuilding from the requested URL mismatches the urlkey when the capture was recorded under a different path form and wayback serves its calendar page instead of the capture. An unreachable archive now answers `transient` with its own message instead of a false `permanent`.
  
2. **Unusable snapshots collapsed into the live error.** Every give-up now carries its exact stage (`structuredContent.archive_stage`): `lookup_unreachable`, `no_snapshot`, `snapshot_fetch_failed`, `snapshot_verdict_rejected`, `snapshot_binary`, `snapshot_extract_thin(N)`. `archive=only` builds its error from the stage; the `auto` path records the stage next to the standing live error so a silent snapshot-side failure is visible in the payload.
  
3. **Thin snapshots were discarded, but a dead domain's last capture is very often a meta-refresh stub** that extracts to zero text and chains to the capture holding the content. Serving is now gated on wayback stub markers, and a stub chain-hops through up to 4 `<meta http-equiv="refresh">` targets when, and only when, wayback rewrote the target (`web.archive.org/web/<14-digit-ts>/...`). Live-web refresh targets are never followed: the URL being resurrected may still be dead, moved, or hostile.
  

## Type

- [x] `feat` — new feature (the transport-dead gate reverses the deliberate refusal to resurrect transport errors; argued above)
- [ ] `fix` — bug fix
- [ ] `test` — tests only

## Checks

- [x] `cargo test --features ocr,rerank` passes (622+ tests) — ran the `ocr,rerank,http` superset via nextest on the `ci` profile (1.98.1, pinned, current master 30b8b34): 1179/1185 green; the 6 failures reproduce on pristine master in a display-less/onnx-less container (xvfb×2, screenshot-path×3, onnx-probe) and are unrelated to this change
- [x] `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` passes (zero warnings) — ran the `ocr,rerank,http` superset
- [ ] CI is green on all 3 platforms (Linux, macOS, Windows) — runs on open

## Breaking changes

None for callers: new behavior lands strictly where the live fetch previously returned a bare error, and `archive=auto`'s documented contract ("dead links") now covers the transport-dead. New structuredContent keys (`fetch_error`, `archive_stage`) are additive. No `feat!` needed.

## Test plan

- 12 new inline tests in `resurrect_tests`:
  - meta-refresh extraction: wayback-rewritten target followed; case-insensitive match with case-preserving target (wayback paths are case-sensitive); unquoted-attribute forms rejected rather than misparsed
  - `wayback_ts_of` validation: malformed timestamps, non-wayback hosts, spoofed `web.archive.org` lookalikes all rejected
  - stub markers: the "Got an HTTP NNN at crawl time / Redirecting to... / Impatient?" interstitial detected; real pages and prose *about* the Wayback Machine not caught
  - `cdx_latest`: nearest-to-present data row wins; header-only, empty, and garbage rows all `None`
  - `transport_class`: death (`tls`, `dns`, `refused`) separated from ambiguity (`timeout`, `reset`, `network`, `protocol`, `ghost`); the resurrectable set is exactly `tls | dns | refused`
  - stage tags: stable machine strings for `structuredContent.archive_stage`
- Full suite green locally on the pinned 1.98.1 toolchain with the `ci` profile, modulo the 6 environmental failures noted above; Windows paths in this PR are `cfg`-free (logic is platform-neutral string/JSON handling), so the `win-check` cross-check risk is nil, but CI confirms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,21 @@ channel until the v4.0.0 release train.
   its value and origin (`--markdown` for the reference table,
   `--legacy` for the old-name map); `donsetch doctor` warns about
   deprecated env names active in the shell.
+- Resurrection reaches transport-dead URLs: a TLS handshake against
+  a parked domain, a dead DNS name and a refused port now gate
+  `archive=auto` resurrection the same way 404/410 do, while
+  timeouts, resets and protocol errors stay excluded (a snapshot
+  must never launder an unknown, and a reset can be an IP-level
+  block that an archived copy would paper over). The gate reads a
+  new `structuredContent.fetch_error` transport class recorded on
+  every status-0 fetch error.
+- Wayback snapshot serving survives wayback's own redirect
+  interstitials: a thin capture is re-checked against wayback stub
+  markers, and a stub chain-hops through up to 4 `<meta
+  http-equiv="refresh">` targets when (and only when) wayback
+  rewrote the target (`web.archive.org/web/<14-digit-ts>/...`); a
+  live-web refresh target is never followed, because the URL we
+  are resurrecting may still be dead, moved, or hostile.
 - `web_screenshot` MCP tool: a rendered PNG of a page through the
   existing tier-2 browser (url, full_page, wait_ms). The capture is
   in-process only; the MCP result carries an image content block and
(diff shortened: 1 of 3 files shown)


dondai44423:
Verified and landed. The policy argument is the strongest part: tls/dns/refused are the three unambiguous forms of a domain being gone, so refusing resurrection on them was leaving the archive unused for the archetypal dead link, while timeouts, resets and protocol errors stay excluded because a snapshot must never launder an unknown. I especially rate the reset/eof sub-classification inside TLS errors, the stage-tagged give-ups (lookup_unreachable vs no_snapshot was a real lie the old code told), the CDX fallback for the scheme-strict availability API, and the stub chain-hop being wayback-only. Shim-verified (forcing TLS resets into the tls class fails your classification test), scoped batch + full gate x2 green (1200 + 1 skip).

PR #226, opened 2026-09-14
feat(byok): add Serply (Google SERP) as a BYOK provider

### LINKED ISSUE #220: feat(search): add Serply as a BYOK search provider
## What

Add Serply (https://serply.io) as a BYOK search provider, wired through
`donsetch keys add serply <key>` like the other cloud providers.

Disclosure up front: I work with Serply, so I have an interest here. I read
CONTRIBUTING's "PRs for new providers or requests for specific ones are both
welcome" as an invitation to ask rather than to just drop a PR, and #94 set the
pattern of asking first, so here is the ask.

## Why

The BYOK list already covers general search (Tavily, Exa), Google SERP
(Serper, SerpApi, SerpBase) and the scraping-backed providers. Serply is
another Google SERP endpoint, but the thing that makes it worth a slot is that
it serves the Google verticals from one key: web, news and scholar. Today
`Intent::Paper` and `Intent::News` only get vertical-aware treatment from
`serpapi.rs`; every other adapter either ignores `intent` or has no vertical to
route to. A second provider that honors the routing gives users a fallback that
does not silently flatten a paper query into plain web results.

It is opt-in behind a user-supplied key and changes nothing about the keyless
default, so it sits in the same "never required for the core toolset" bucket as
the existing nine.

## How

Mirror `serpapi.rs` rather than `serpbase.rs`, since that is the adapter that
already routes by intent:

- `GET /v1/search` with the key in an `X-Api-Key` header, so it never lands in a
  logged request line. Hits come back in `results[]` with `position`, so the
  existing `1.0 / position` score derivation applies unchanged, and `num` is
  capped at 10 like the other adapters.
- `Intent::News` adds `tbm=nws` to the same endpoint. That keeps the `results[]`
  shape and real publisher links. (There is a `/v1/news` RSS endpoint, but it
  ignores `num` and its links are news.google.com redirects, so it is the wrong
  route for this.)
- `Intent::Paper` goes to `GET /v1/scholar`, which returns `articles[]`.
- Key rotation classification the same way the others do it: 401/403 invalid,
  402 credit depleted, 429 rate limited or credit depleted by message, 5xx
  server error.

Two shape details worth calling out because they are easy to get wrong: the
snippet field is `description`, not `snippet`, and scholar entries carry no
`position` field while the scholar envelope also ships an empty `results: []`
next to `articles[]`. A parser left pointed at `results` returns zero hits and
looks like an empty search rather than a bug, so the results key has to be
passed in explicitly and array order has to stand in for the missing position.

I have the adapter written and gated locally already: 5 files (+262/-2), same
footprint as #109, five unit tests on the pure parser, `cargo fmt --check`
clean, `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` clean,
and the lib suite at the same pass/fail counts as an unmodified checkout. I
also ran it end to end against a live key: all five intents return hits through
`donsetch search`, and a bad key gets marked dead and falls back to local.

Happy to send the PR whenever you want it, or to adjust the design first if you
would rather it went another way. No pressure either way.

googio:
## Summary

Adds Serply (https://serply.io) as another BYOK Google SERP provider, per the discussion on #220. It follows the existing `serpapi` adapter rather than `serpbase`, because Serply serves the Google verticals and the repo already routes by `Intent`: paper goes to the scholar endpoint, news to `tbm=nws`, everything else to plain web search. One adapter file plus the usual registry, help and README lines.

## Type

- [x] `feat` new feature

## Checks

- [ ] `cargo test --features ocr,rerank` passes (622+ tests)

Two boxes left unticked rather than overclaimed:

- The suite is 1208 passed / 25 failed on this branch. The same 25 fail on an unmodified `v4.1.0` checkout on this machine, so they are not from this change, but the command still exits non-zero and I am not going to tick a box saying it passes. Numbers and the base comparison are in the test plan.
- CI on this PR needs a maintainer to approve the fork run, so I cannot observe the 3-platform result myself.

## Breaking changes

None. `PROVIDERS` gains an entry and `dispatch` gains an arm. Nothing changes for anyone who does not run `donsetch keys add serply`.

## Changes

- New: `src/search/byok/serply.rs`, a `search()` adapter plus a pure `parse_results()` (5 unit tests)
- Updated: `src/search/byok/mod.rs`, module declaration and dispatch arm
- Updated: `src/search/byok/store.rs`, provider registry and the abbreviated list hint
- Updated: `src/cli/keys.rs`, module doc list and the `keys add` help line
- Updated: `README.md`, the `keys add serply` example

No new dependencies. `CHANGELOG.md` is left alone, the way #109 left it.

## API contract notes

Measured against the live API, not read off the docs (https://serply.io/docs).

- `GET https://api.serply.io/v1/search`, params `q` and `num`, auth via the `X-Api-Key` header so the key never lands in a logged request line
- Paper intent uses `GET /v1/scholar`, which returns its hits under `articles[]` and ships an empty `results: []` next to them. A parser left pointed at `results` returns zero hits and looks like a successful empty search, so the array name is passed in explicitly and there is a test for exactly that
- News adds `tbm=nws` to the search endpoint. That keeps the `results[]` shape and real publisher links, so news needs no second parser. `/v1/news` also exists but is an RSS feed: it ignores `num` and its links are `news.google.com` redirects rather than the article
- Results map `position`, `title`, `link`, `description` to `SearchHit` with a position-derived score, same as serpapi and serper. The snippet field is `description`, not `snippet`. Scholar entries carry no `position`, so array order stands in
- HTTP 401 and 403 map to `InvalidKey`, 402 to `CreditDepleted`, 429 to `CreditDepleted` or `RateLimited` by message, 5xx to `ServerError`. Some error envelopes arrive as HTTP 200 with a `detail` string instead of a status, so that is checked too and a key-related `detail` maps to `InvalidKey` so rotation marks the key dead
- `num` is capped at 10 to match the other providers, and the result list is truncated locally as well, since the cap is approximate on some verticals

## Test plan

Rebased onto `v4.1.0` (`fec3e83`) and re-run on this machine, Linux, Rust 1.98.1.

Unit tests, 5 new ones on the pure parser covering web mapping, the scholar `articles` key next to an empty `results`, dropping entries without a link, empty on a missing key, and the array-order fallback when `position` is absent:

```
cargo test --features ocr,rerank --lib serply -- --test-threads=1
test result: ok. 5 passed; 0 failed; 0 ignored; 1229 filtered out
```

Full suite, this branch against an unmodified `v4.1.0` checkout, same command and same machine:

```
base   fec3e83 : 1203 passed; 25 failed; 1 ignored
branch          : 1208 passed; 25 failed; 1 ignored
```

Exactly the 5 new tests, and the failing set is identical on both sides (I diffed the two lists). Those 25 are environment dependent here, xvfb, kill-switch env vars and filesystem or network state, and none is in `src/search/byok/`. `--test-threads=1` matters, several of them mutate process env and race at the default parallelism.

Lint and format:

```
cargo fmt --all -- --check                                      clean
cargo clippy --all-targets --features ocr,rerank -- -Dwarnings  zero warnings
```

End to end against the real API, with `DONSETCH_CACHE_DIR` pointed at a scratch dir so the key did not persist:

```
donsetch keys add serply <key>
donsetch search "postgres vacuum tuning" --intent <intent>

web     7 hits  2799ms
code    7 hits   289ms
entity  7 hits   727ms
news    7 hits  3087ms
paper   7 hits  4029ms
```

News came back with publisher links (reuters.com, anthropic.com) rather than Google redirects, and paper with doi.org links carrying author and journal lines. Key rotation classification checked separately: a deliberately bad key flipped to `invalid` in `keys list` after one search, and search fell back to local.

Closes #220.

Serply support is optional; behavior is unchanged when no `serply` key is added.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this entirely if it isn't a direction you want for the project. On the plugin question @Mart-Bogdan raised in #220, the in-tree versus plugin call is yours; say the word and I will close this and publish a standalone plugin adapter instead.


googio:
Understood, and thanks for the quick answer. Closing this, as promised in the description.

That is a reasonable bar to hold. A short, deliberate list is easier to defend than one that grows a vendor at a time, and the plugin system genuinely does cover this case.

I will put together a Serply plugin instead. One thing that would make that path better for every plugin author, independent of whether Serply is ever in tree: @Mart-Bogdan offered in #220 to expand the plugin API with better error code support, and there is a concrete gap. `KeyError` in `src/search/byok/mod.rs` carries six variants that drive different key-state transitions, `InvalidKey` retiring a key permanently while `RateLimited` is a cooldown, but the plugin contract returns a single `retryable` boolean. A plugin therefore cannot tell the store that a key is dead rather than throttled, so a bad key keeps getting retried. If you want that closed, I am happy to send it as a separate PR with no vendor content in it.

Thanks for looking at it.

PR #239, opened 2026-09-18
docs(cargo): say which features CI and release actually build with

Mart-Bogdan:
## Summary

The [features] comment claimed CI and release builds enable `--features ocr,rerank`. Release and most CI lanes have added `http` since then, and some targets build with fewer features or none (linux-aarch64, macos-x86_64). Point at the workflow matrices instead of restating them, and say what `http` is for.

## Type

- [ ] `feat` — new feature
- [ ] `fix` — bug fix
- [x] `docs` — documentation only
- [ ] `test` — tests only

## Checks

Just comment change. I didn't run anything

## Breaking changes

Just comment change.

## Test plan

Just comment change.


dondai44423:
Thanks. The red earlier was my test, not your change: the daemon-restart test raced `sh` startup on the macos-x86_64 runner and took that lane down twice on a doc-only diff. Fixed on master (the write now waits for the child to close its own stdin instead of trusting a 300ms timer), and your rebased head is green on all five platforms.

Checked the text against both matrices before merging: CI builds linux-x86_64 with ocr,rerank,http, macos-arm64 with ocr,rerank, and the other three with none; the release builds linux-x64, darwin-arm64 and win32-x64 with the full set, linux-arm64 and darwin-x64 with none. `http` for the axum transport is right too.

PR #254, opened 2026-09-19
feat(plugins): let a search plugin report why it failed

googio:
## Summary

A search plugin can now say **why** it failed, and DonSeTch acts on it. The format-1 error envelope gains an optional `error_kind` (`invalid_key`, `credit_depleted`, `rate_limited`, `server_error`, `network_error`), which is classified into the same `KeyError` variants a native adapter derives from an HTTP status, and which then drives plugin selection the way a native key's state drives key selection.

This is @Mart-Bogdan's suggestion on #220, "We can expand plugin API with better error code support, with some additional fields in responses etc." and "Better to improve plugin API if there is something unsupported." No vendor content, as promised on #226.

## Type

- [x] `feat` new feature

## Why

Today a plugin has one bit, `retryable`, and even that bit is thrown away:

- `parse_envelope` never reads `retryable` at all. Its error branch builds a `String`.
- `run_plugin`'s non-zero-exit branch reads it and then discards it: `if let Ok((messages, _)) = extract_error_envelope(&body)`.
- Every plugin failure therefore arrives as `KeyError::UnknownError`, whose `to_key_state()` is `None`.

So a plugin whose API key was revoked at 09:00 is spawned again on every search for the rest of the day, each one a process spawn plus a round trip to an upstream that will answer 401 again.

Carrying the classification across the boundary would not have been enough on its own, which is the part worth reviewing:

- `ByokConfig::update_key_state` (`store.rs:415`) returns early for a provider that is not in `self.providers`, and a plugin never is.
- `pick_any_skipping` (`mod.rs`) selects plugins without consulting any state.

A protocol-only change would have added a field that nothing could read. So `PluginDef` carries the `state` and `ts` a `KeyEntry` carries, and the picker honours them.

## The contract

```json
{"format":1,"error":"401 Unauthorized: API key revoked","error_kind":"invalid_key"}
```

| `error_kind` | `KeyError` | Effect |
|---|---|---|
| `invalid_key` | `InvalidKey` | parked until the user re-registers the plugin |
| `credit_depleted` | `CreditDepleted` | parked until the user re-registers the plugin |
| `rate_limited` | `RateLimited` | parked for the 60s `RATE_LIMIT_COOLDOWN`, then auto-recovers |
| `server_error` | `ServerError` | transient, no state change, next provider |
| `network_error` | `NetworkError` | transient, no state change, next provider |
| absent, or a value this build does not know | falls back to `retryable` | no state change, exactly as today |

Recovery for the two sticky states is `donsetch keys add plugin <name> --cmd '...'` again, which is what a user does anyway after fixing the credentials the plugin uses. `PluginConfig::add` resets the state, and that is in the README line.

## Breaking changes

None, and it is worth being precise about why rather than just asserting it.

- A plugin that sends no `error_kind` gets `ServerError` or `UnknownError` from `retryable`, both of which are `to_key_state() == None`, so no state is ever recorded. Its behaviour is byte-identical. There is a unit test asserting exactly that, and a live probe below.
- A `plugins.json` written by an older build has no `state` or `ts`. Both are `#[serde(default)]`, so it loads as `active` with no migration.
- `keys list` prints an active plugin exactly as before. The state only appears once it is not active.

## Changes

- `src/search/byok/plugin.rs` - module doc for the extended envelope; `PluginDef.state` / `.ts` with serde defaults and a `Default` impl; `PluginConfig::mark_state` / `take_usable`; a `PluginError` type that both envelope readers now share, replacing a duplicated parse; `parse_envelope` returns `Result<_, KeyError>`; `PluginStore::mark_state` / `is_usable` (lock, mutate, persist).
- `src/search/byok/mod.rs` - `pick_any_skipping` skips a parked plugin at both of its plugin-selection points; the `Err(key_error)` arm routes to the plugin store when the provider was a plugin.
- `src/search/byok/store.rs` - `RATE_LIMIT_COOLDOWN` and `now_ts` widened from private to `pub(super)`, so a throttled plugin waits exactly as long as a throttled native key rather than growing a second number.
- `src/cli/keys.rs` - the plugin block renders `def.state` with the icon vocabulary `render_list` already uses, instead of a hardcoded green tick.
- `README.md` - one paragraph in "Search plugins".

No new dependencies, `Cargo.lock` untouched.

## Test plan

**Unit, 7 new in `search::byok::plugin::tests`:** the five `error_kind` values mapping to their key states; case and whitespace tolerance; an unknown kind falling back to `retryable`; a legacy envelope recording no state either way round; the state machine including cooldown expiry and an unregistered name; re-registration reviving a parked plugin; and a `PluginDef` deserialized without `state`/`ts` loading active.

**Live, against `target/fast/donsetch` with `DONSETCH_CACHE_DIR` pointed at a throwaway directory.** Four plugins, each a `/bin/sh` script that appends to a log and prints one envelope:

| Plugin | Envelope | 3 searches | State after | `keys list` |
|---|---|---|---|---|
| `counted` | `error_kind: invalid_key` | **1 spawn** | `invalid` | `✗ ... invalid` |
| `throttled` | `error_kind: rate_limited` | **1 spawn** | `rate_limited` | `⏱ ... rate-limited` |
| `throttled`, `ts` wound back 61s | same | 1 more spawn | `active` | `✓ ...` |
| `legacyplug` | `retryable: true`, no kind | **3 spawns** | `active` | `✓ ...` |

The legacy row is the compatibility proof: 3 spawns from 3 searches is what every plugin in existence does today, and still does. The first row is the change: 1 spawn instead of 3. Re-registering `counted` put it back to `active` and `✓`.

Each search fell through to the keyless engines and returned results, so the fallback chain is intact with a parked plugin in it.

**Gates**, all on this branch:

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets --features ocr,rerank -- -Dwarnings` | clean |
| `cargo clippy --all-targets --features ocr,rerank,http -- -Dwarnings` | clean |
| `cargo check --locked --profile ci --all-targets --features ocr,rerank,http` | clean (the `lockgate` recipe) |
| `cargo test --features ocr,rerank --lib -- --test-threads=1` | 1245 passed / 26 failed |

The 26 failures are identical on a detached checkout of this PR's base, `25f9578`, which is 1238 passed / 26 failed: the same 26 names, and the 7 extra passes are the new tests. They are environment-dependent on this machine (xvfb, cache paths, `routes.json`), not from this change. I have left the test box unticked rather than tick a box for a command that exits non-zero.

## Checks

- [ ] `cargo test --features ocr,rerank` passes (see the base comparison above)

## Size

432 lines across 5 files, over the size I would normally aim for, so here is the split: 121 of the 360 added lines are tests, and roughly 44 more are doc comments and the module header. The functional change is the `PluginError` type, two `PluginConfig` methods, two `PluginStore` wrappers, three call sites in `mod.rs` and one render block in `keys.rs`.

The 10 `..PluginDef::default()` lines in the existing `run_plugin` tests are the cost of adding a field to a public struct that tests build by literal. Using the spread rather than spelling out both fields means the next field added to `PluginDef` will not touch them again.

Happy to split the state persistence out of the protocol change if you would rather review them separately, though as described under **Why** the protocol half does nothing on its own.

## Notes and disclosures

- **The plugin's own message is dropped for the three parked kinds.** `InvalidKey`, `CreditDepleted` and `RateLimited` carry no payload, so `"401 Unauthorized: API key revoked"` becomes `"invalid key"` on the model surface. That is what every native adapter already does for 401/402/429, so it is house-consistent rather than a new loss, and `keys list` names the state. Say the word if you would rather those variants carried a string.
- **One message change on the non-zero-exit path.** That branch returned the envelope's text with no prefix; it now reads `plugin <name>: <text>`, matching the zero-exit path, which always had the prefix.
- **`keys reset` deliberately not extended.** It validates its argument against keyed providers and exits 1 when no keys exist, so teaching it plugins is its own change. Re-registering is the recovery path and it is documented. Happy to add it if you want the symmetry.
- **`CHANGELOG.md` untouched.** #252 is currently adding an `[Unreleased]` section that does not exist on `master` yet; a second PR creating the same heading would conflict for no reason. Tell me and I will add an entry, before or after #252 lands.
- **#252 touches `src/cli/keys.rs` and `src/search/byok/store.rs` too**, in different hunks (its legend and help strings, my plugin render block and two visibility keywords). I do not expect a conflict, and I will rebase behind it either way.
- **`just win-check` was not run**: this box has no mingw cross toolchain and no sudo. The two `cfg(windows)` `PluginDef` literals in the tests got the same one-line spread as the eight unix ones, so I expect it clean, but I have not proven it.
- **On whether this was wanted.** @Mart-Bogdan named this gap on #220 and I offered it twice, on #220 and on #226, with no reply since 09-15. I read the offers as real and built it, but nobody said "send it" in those words, so if the answer is no, close it and I will not push back.


Mart-Bogdan (changes requested):
Thanks for this. It's what I had in mind on #220, and the write-up made the review easy. A few things before it lands:

**1. `keys add plugin --test` loses the plugin's own error text.** You flagged the message loss for search results, but it hurts most in the probe. `probe()` returns `e.to_string()`, so a plugin that answers `invalid_key` prints `probe failed: invalid key`. The plugin name and its "401 Unauthorized: API key revoked" are both gone, and that's exactly when the user is debugging their credentials. Please keep the plugin's text on this path. Either let `InvalidKey` / `CreditDepleted` / `RateLimited` carry a message, or keep the envelope text for the probe and the debug log.

**2. Tests: the parser is covered, the behavior isn't.** The new tests cover both envelope shapes at the `parse_envelope` level, plus the in-memory state machine. The two parts that change behavior have no test:

- **The non-zero-exit path in `run_plugin`.** The `Ok(code)` branch → `extract_error_envelope` → `into_key_error` is the second place an envelope is read, and nothing tests it with either shape. Two `/bin/sh` tests in the style of `failer` would cover it. One prints `{"format":1,"error":"x","error_kind":"invalid_key"}` and exits 3, and asserts `to_key_state() == Some(KeyState::Invalid)`. The other prints the same envelope without `error_kind` and asserts no state and that the plugin's text survives.
- **`ByokSearcher::search` end to end.** The routing in `mod.rs` (a plugin error goes to `plugins.mark_state`, not `store.update_key_state`) and the new skip in `pick_any_skipping` are only covered by your live run. The repo already has the pattern for this: `DONSETCH_CACHE_DIR` is read on every call, and `engine_health_persists_across_restart` (`src/search/mod.rs`) points it at a throwaway directory, which is safe because nextest runs each test in its own process. Register a `/bin/sh` plugin that appends to a counter file and answers `invalid_key`. Search twice, then assert one spawn and `invalid` in `plugins.json`. Repeat with the legacy envelope and assert two spawns and `active`. That also covers the reload-from-disk at the start of every search, which the whole design relies on. It would be your live table as a test. This one can be a follow-up PR if you'd rather not grow this one. The non-zero-exit tests above I'd like here.

**3. `doctor` reports a parked plugin as healthy.** `check_plugins` only looks at registration, so a plugin sitting in `invalid` still passes. Returning `Warn` with the state for any plugin that isn't `active` would close the loop, since otherwise `keys list` is the only place it shows. Fine here or as a follow-up, your call.

Two notes, no action needed:

- **The message text isn't byte-identical.** A legacy `retryable:true` envelope now becomes `ServerError` instead of `UnknownError`, so the text changes from `plugin x: boom` to `server error: plugin x: boom`. That's fine, but when the changelog entry is written it shouldn't say legacy plugins are unchanged. The *state* behavior is unchanged, the wording isn't.
- **The daemon now writes `plugins.json`.** Before this PR only the CLI did. A `mark_state` save at the end of a plugin run (up to 30s) can overwrite a `keys add/remove plugin` made in that window. `byok-keys.json` already has the same race, so I'm just noting that it now reaches one more file.

On `just win-check`: CI is green on all three platforms, so it's fine for this PR. It matters more for @dondai44423, who pushes directly to `master` without a CI gate in front. Do you have Docker available? If we provided a Docker path for running `win-check`, would you be able to use it?

Issue #261, opened 2026-09-19
Please re-enable tests on CI on windows


Mart-Bogdan:
PR #254 introduced test that was failing on Windows, but there were no way to detect this without CI. I have fixed it in follow-up PR #260

But that was because I was right here when it happened, and I decided to check what PR changes was after my review.

This especially important for Windows target, as it differ from Linux even more than macOS,

We should at least then run tests on release pipeline so it would fail loudly before release, so it can be fixed, but I'd prefer tests to be green on `master` as well.

If issue is performance of tests -- perhasp we can split tests into 2-3 batcehs and run them on separate jobs (runners) in parallel?

Also I have noticed that `cargo test` runs slingltly faster on windows (no need to spawn new sub-proceses), but `cargo tests` don't work now as many tests rellie on environment change or global state, so can't be parllelized. So only part of tests can be run with `cargo test`.

I'd prefer test on macOs and Windows to be enabled back. and perhaps in future split tests by modules and run as separate parallel steps. This would solve performance issue I think.

Mart-Bogdan:
Follow-up evidence from today: the v4.2.6 release run failed on win32-x64, darwin-arm64 and darwin-x64 (Linux passed, "Create Release" was skipped). All three hit the same compile error from `8e48dc8`:

```
error[E0609]: no field `child` on type `other::Xvfb`
   --> src/ghost/manager.rs:210:33
```

This one is a slightly different gap from the one above. Windows CI does still compile-check every target, so it would have caught it, but the commit went straight to `master` and the release was tagged before CI on it finished. So besides re-enabling tests, it would help if `release.yml` refused to build unless CI on the tagged commit was green (or ran the same compile gate first).

On splitting the suite: nextest supports this directly with `cargo nextest run --partition count:1/3` (`2/3`, `3/3`) across parallel matrix jobs. There's no module list to maintain, and it keeps the one-process-per-test isolation that the `DONSETCH_CACHE_DIR` tests rely on.

```
cargo nextest run --partition count:1/3
cargo nextest run --partition count:2/3
cargo nextest run --partition count:3/3
```

dondai44423:
Fixed in 4.2.6, and you were right on both counts.

[...]

PR #262, opened 2026-09-19
test: build the integration tests as one binary

Mart-Bogdan:
## Summary

The ten files directly under `tests/` were each their own crate, so each one linked the whole library and dependency tree again, and re-generated the generic code it used from the library. This moves them into `tests/it/` as modules of one crate (`tests/it/main.rs`), so they build and link once. Related to #261: while measuring where Windows CI time goes, the test build turned out to cost far more than running the tests (18 s of runtime against a ~13-minute test step).

## Type

- [x] `test` — tests only

## Measurements

These are build-only timings (`cargo nextest run --no-run --profile ci --cargo-profile ci --timings`), default features, dependencies already compiled, `donsetch` cleaned before each run. Local machine, 24 threads, so a 4-core runner will differ.

| | Before (10 test binaries) | After (1 `it` binary) |
|---|---|---|
| opt-level 3 (the `ci` profile today) | ~93 s combined across the 10 binaries | **8.7 s** |
| opt-level 0 | 10 × ~1.8 s, in parallel | 2.5 s |

At opt-level 3 the saving is real. At opt-level 0 the integration tests were already cheap and stay cheap. The whole-build time also dropped (95.8 s → 67.2 s at opt-level 3), but library units the change doesn't touch got faster too, so part of that is run-to-run noise. The table claims only the integration-test part.

Before (opt-level 3):

image

After (opt-level 3):

image

## What doesn't change

- **Isolation.** nextest still runs every test in its own process, so the `DONSETCH_CACHE_DIR` tests are unaffected.
- **Filters.** Test names gain a module prefix (`soak::soak_rss_stays_bounded`, `egress_proxy::…`). Every existing filter matches on test names and selects the same tests as before: the heavy set `test(soak) | test(corpus) | test(landmark)`, aarch64's `not test(pdf)`, and the macOS smoke list. Nothing uses `binary(…)`.
- **Paths.** No test used crate-level attributes, `crate::` paths or relative `include_str!`. The two `env!` uses (`CARGO_MANIFEST_DIR`, `CARGO_BIN_EXE_donsetch`) work from any integration-test crate.

One thing for future contributors: a new integration test goes in `tests/it/` with a `mod` line in `tests/it/main.rs`. A new top-level `tests/*.rs` file still works, but quietly becomes its own binary again. The CHANGELOG entry says so.

## Checks

- [x] Tests pass under nextest (the suite requires it), on the `fast` profile, Windows:
  - `ocr,rerank,http`: 1,299 run, 1,298 passed, including all 41 `it` tests.
  - default features: 1,270 run, 1,269 passed, including all 41 `it` tests.
  - The single failure is `handles::tests::is_handle_recognizes_and_rejects` in both runs. It reads the machine's real `donsetch.toml` and fails the same way on `master`.
- [x] `cargo clippy --profile ci --all-targets -- -Dwarnings` passes, both with default features and with `--features ocr,rerank,http`
- [x] CI is green on all 3 platforms (Linux, macOS, Windows)

## Breaking changes

None. Test layout only. `src/` changes by one comment path (`tests/it/token_invariants.rs`).

## Test plan

Per-module pass counts in the `it` binary match the number of test functions in each original file: auth_login 14, bypass_live 6, egress_proxy 6, token_invariants 6, request_class 4, revalidate_redirect 2, crawl_fresh_fetch 1, daemon_boot_stays_alive 1, secure_cookie_leak 1. `soak` is `#[cfg(target_os = "linux")]`, so it runs only on Linux, as before.


Mart-Bogdan:
CI numbers, since the local timings in the description came from a 24-thread machine. Same code on both sides: the baseline is `master`'s own run for 9ef1226, which is this PR's base, and I triggered a second PR run to get two samples. Every run restored its cache with `full match: true` and recompiled only `donsetch`, so this is the `donsetch` + test-binary build and nothing else.

Build time of the test step (the `Finished` line):

| Lane | `master` @ 9ef1226 | PR run 1 | PR run 2 | Δ vs PR avg |
|---|---|---|---|---|
| linux-x86_64 | 4m 15s | 3m 57s | 4m 06s | −14s (−5%) |
| linux-aarch64 | 2m 43s | 2m 35s | 2m 36s | −8s (−5%) |
| macos-arm64 | 5m 36s | 5m 24s | 3m 31s | runner noise |

- **Linux:** both PR samples beat the baseline on both architectures, a modest but consistent gain. A 4-core runner with Linux's linker saves far less than the local chart suggests, because `donsetch "lib" (test)` stays on the critical path either way.
- **macOS:** the two PR samples differ by almost 2 minutes on identical code, so runner variance swamps the effect there.
- **Test counts are identical** on every lane, before and after: 1346 on linux-x86_64, 1274 on linux-aarch64, 437 on the macOS smoke set, all passing. So the move didn't drop or duplicate any test.

The number I'd most like to see is Windows, once this reaches `master`. PRs only compile-check that lane, so it can't show up here. Since 9ef1226 the push run builds and runs the suite there, and the ten separate test binaries were exactly the part MSVC pays for per binary. For reference, 9ef1226's Windows test step took 12m 28s to build (run `35459270025`). The first `master` run after merge gives the before/after on the lane that motivated #261.

Mart-Bogdan:
Correction to my numbers above. #263 gave me a second "before" sample — same base 9ef1226, only `src/mcp/compat.rs` and the changelog changed, so it still builds the ten test binaries — and it does not support the aarch64 or macOS figures.

| Lane | before: master @ 9ef1226 | before: #263 | after: #262 run 1 | after: #262 run 2 |
|---|---|---|---|---|
| linux-x86_64 | 4m 15s | 4m 14s | 3m 57s | 4m 06s |
| linux-aarch64 | 2m 43s | 2m 33s | 2m 35s | 2m 36s |
| macos-arm64 | 5m 36s | 4m 08s | 5m 24s | 3m 31s |

- **linux-x86_64 holds**: two before-samples 1s apart, both after-samples below them. ~13s, about 5% of the step.
- **linux-aarch64 does not.** #263 built in 2m 33s, faster than either run of this PR. My earlier −8s was one slow baseline, not an effect.
- **macos-arm64 is noise**, now 88s of spread across before-samples alone.

All six runs restored with `full match: true` and recompiled only `donsetch`, and test counts are unchanged everywhere (1346 / 1274 / 437, all passing).

So on PR lanes the honest claim is ~5% on linux-x86_64 and nothing measurable elsewhere. The local numbers stand on their own: the ten binaries' compile work is gone, which matters most where cores are few and per-binary link cost is high. Windows is still the lane to watch after merge.

Mart-Bogdan:
Better local numbers, because the first pair was thermally skewed. My machine hit 98 °C and throttled during those benchmarks, and the "before" run was the fourth consecutive build in an opt-level sweep, so it ran at a lower clock than the "after" run. That inflated the difference: library units the PR never touches appeared 34% faster in the "after" run, which was the tell.

Re-measured with the fan pinned and both runs started from a settled temperature, otherwise identical (opt-level 3, `ci` profile, default features, deps warm, `cargo clean -p donsetch` before each, build only):

| | master @ 9ef1226 | this PR |
|---|---|---|
| wall time | 1m 13s | 1m 14s |
| integration-test units | 10 units, **81.5s combined** | 1 unit, **10.1s** |
| all `donsetch` units | **277 unit-seconds** | **198 unit-seconds** |

So the honest claim is narrower than my first one: **wall time on a 24-thread box is unchanged; what goes away is ~29% of the build's compile work.** The ten test binaries were never on the critical path here — they compile in parallel with `donsetch "bin"`, which is the real tail — so on a machine with spare cores the saving is invisible. The `--timings` charts show it plainly: on master the last 20 seconds runs 14 units with the CPU band near the top, on this branch the same window is `donsetch "bin"` alone at much lower utilisation.

Where cores are scarce that work becomes wall time, which is exactly the ~5% measured on linux-x86_64 CI. And it says the 93s figure in the PR description was inflated by heat; the real number is 81.5s → 10.1s.

That also means Windows remains the only lane where this could matter much (ten MSVC links, a ~12.5 minute test step), and PRs can't show it. I can run master and this branch through my fork, where a push runs the Windows suite, if that number would help the decision.

## Before this patch:
image

## After this patch:
image

From charts we can see that time is ~same but CPU usage is lower on tail part.

Mart-Bogdan:
Windows numbers, as offered above. Dispatched `master` and this branch manually, since `workflow_dispatch` runs the non-PR steps and therefore builds and runs the suite on that lane. Each pair was launched together so both sides share time-of-day and runner pool. Times are the `Test (nextest, Windows core suite, non-PR)` step; 1270 tests passed in every run.

| Pair | master | this branch |
|---|---|---|
| 1 ([master](https://github.com/dondai44423/donsetch/actions/runs/35476104808) / [branch](https://github.com/dondai44423/donsetch/actions/runs/35476111299)) | 13m 00s | 9m 15s |
| 2 ([master](https://github.com/dondai44423/donsetch/actions/runs/35477188290) / [branch](https://github.com/dondai44423/donsetch/actions/runs/35477190854)) | 13m 07s | 12m 32s |
| warm cache, my fork ([master](https://github.com/Mart-Bogdan/donsetch/actions/runs/35477312122) / [branch](https://github.com/Mart-Bogdan/donsetch/actions/runs/35477457317)) | 5m 38s | 5m 06s |

Pair 1's 9m 15s is runner luck rather than the change: `master` is stable at 13m 00s / 13m 07s, while this branch produced 9m 15s and 12m 32s on identical code. Had I stopped at pair 1 I would have reported a 29% win.

The fork pair is the one to trust. Both sides there restored a cache that actually contains compiled code — 5 crates recompiled, against 223 on upstream — so the step measures the `donsetch` build and little else. The gap is **32s, about 9%**.

That squares with everything else: ~0% on a 24-thread desktop, ~5% on linux-x86_64, ~9% on a warm Windows lane. It's a modest win that grows as cores get scarcer, and I'd rather state it that way than at its most flattering.

The 223-versus-5 crates in that paragraph is a separate and much larger finding about the Windows lane's cache, worth roughly 7.5 minutes per run. It has nothing to do with this PR, so it goes on #261.

PR #264, opened 2026-09-20
ci: key the cargo cache by what the job builds, not by what the matrix declares

Mart-Bogdan:
## Summary

The Windows lane's cache is warm for every step except the one that now dominates the job. `ci.yml` keys the cache by `matrix.target`, but the Windows lane builds two different shapes under that one key: `cargo check` on pull requests, and the full suite on every other event since 9ef1226. A check run leaves type metadata and no compiled code, and `rust-cache` does not re-save when the key hits exactly ("Cache up-to-date."), so the check-shaped entry wins forever — until `Cargo.lock` or the toolchain moves `DEP_HASH`.

The result is that every Windows test run restores a cache, logs `full match: true`, and then recompiles **223 crates** before it can build a single test.

This adds a `CACHE_KIND` derived from what the job actually builds and puts it in the key.

## Type

- [x] `chore` — deps, CI, tooling

## Measurement

My fork had no cache at all, so its first run saved an entry containing both check and codegen artifacts. Later runs there restore it. Both rows are `master` at 9ef1226, both `full match: true`, both 1270 tests passed:

| | upstream ([35476104808](https://github.com/dondai44423/donsetch/actions/runs/35476104808)) | fork, test-shaped cache ([35477312122](https://github.com/Mart-Bogdan/donsetch/actions/runs/35477312122)) |
|---|---|---|
| crates recompiled | 223 | 5 |
| Compile-check all targets | 4m 12s | 32s |
| **Test step** | **13m 00s** | **5m 38s** |
| Compile-gate release features | 21s | 18s |
| Clippy | 20s | 16s |
| **whole job** | **20m 15s** | **8m 29s** |

A genuinely cold fork run took 12m 57s in the test step, against upstream's "warm" 13m 00s — the cache currently does nothing at all for that step.

## Why not just bump the suffix

`-ci1` → `-ci2` would clear today's stale entries, in the style of the earlier `v2 → v3 → v4` bumps. But the Windows lane would still put both shapes under one key, and PR runs are the frequent ones, so the first PR to save would re-poison it within a day. Keying by shape is what prevents the recurrence.

`matrix.tests` alone isn't enough either, for the same reason: it says `none` for Windows on every event. Hence the small computed step, which deliberately mirrors the `if:` on "Test (nextest, Windows core suite, non-PR)" and says so in a comment. If you'd rather the matrix itself told the truth — Windows declared `tests: full` with the event condition moved onto the step — I'll rework it that way; that's a bigger change to how the lane is declared, so I didn't assume it.

## Effect on other lanes

Both Linux lanes and macos-arm64 have always built tests, so they're already correct and only get a one-time key change. macos-x86_64 never builds tests, so nothing changes for it in practice. Every lane pays one cold run after merge, then stays warm.

## Checks

- [x] `ci.yml` parses (`yaml.safe_load`)
- [ ] CI is green on all 3 platforms — this PR's own run is the test: the Windows lane should recompile everything once (new key) and the run after it should recompile ~5 crates

Relates to: #261


Mart-Bogdan:
I have dispatched new workflow https://github.com/dondai44423/donsetch/actions/runs/35479057570 but this would take ~30 minutes as it would have cold cache. 
I probably won't be able to run second workflow as I am going to bed.

Second manually dispatched workflow from branch `ci/cache-for-windows-tests` would show whether cache indeed speed up ci for this windows step.

If I won't run second run, please run it manually:

```
gh workflow run CI --repo dondai44423/donsetch --ref ci/cache-for-windows-tests
```

Mart-Bogdan:
Both halves measured on this branch. The key change works exactly as intended: the first non-PR run is cold (new key), the next one restores it and compiles only `donsetch`.

Upstream, `workflow_dispatch` on this branch, Windows lane:

| | `master` today | this branch, run 1 — cold ([35479057570](https://github.com/dondai44423/donsetch/actions/runs/35479057570)) | this branch, run 2 — warm ([35480541437](https://github.com/dondai44423/donsetch/actions/runs/35480541437)) |
|---|---|---|---|
| crates recompiled | 223 | 418 | **5** |
| Compile-check all targets | 4m 12s | 7m 18s | **31s** |
| **Test (nextest, Windows core suite)** | **13m 00s** | 12m 33s | **6m 27s** |
| Compile-gate release features | 21s | 4m 53s | 19s |
| Clippy | 20s | 4m 23s | 20s |
| **whole job** | **20m 15s** | 30m 55s | **9m 27s** |

Same sequence on my fork, independently ([cold](https://github.com/Mart-Bogdan/donsetch/actions/runs/35479086044), [warm](https://github.com/Mart-Bogdan/donsetch/actions/runs/35480302285)): 417 crates → 5, whole job 27m 47s → 11m 34s.

All runs: 1270 tests passed.

The `master` column is today's behaviour, and the point of the PR: that run restores a cache, reports `full match: true`, and still recompiles 223 crates, because the entry it restores was written by `cargo check` runs and holds no compiled code.

Two honest caveats:

- **One cold run per lane after merge.** Every key changes, so the first run on `master` pays it once, and PR branches pay once more for their own `KIND=none` entry. After that both shapes stay warm independently.
- **Windows runner variance is ±2 minutes even warm.** My fork's warm test step was 7m 52s here and 5m 38s / 5m 06s in an earlier pair on `master`. So the test step's honest range is roughly 5–8 minutes warm against 12.5–13 minutes today. The crate count and the non-test steps are the steady signal: 223 → 5, and check/gate/clippy from 4m 12s / 21s / 20s at 12+ minutes of cold work down to 31s / 19s / 20s.

PR #274, opened 2026-09-20
fix(bypass,proxy): the unlocker response is capped at 64 MiB; a bad proxy entry's error no longer echoes its password

mnaza:
Two small hardening items from a sweep of `transport/proxy.rs` and `fetch/bypass.rs` at 4.2.9. The rest of that sweep was clean: CONNECT head capped at 4 KiB with a timeout, SOCKS5 lengths gated before the `as u8`, `Proxy-Authorization` only on the proxy hop, bracketed v6 proxy hosts stripped at parse (not the #267 bug), the unlocker client has a timeout and default TLS verification, the token appears only in the header.

## 1. Unlocker response uncapped (MEDIUM-LOW, memory; opt-in surface)

`bypass.rs:682-685` and the retry at `:706-709` call `resp.bytes()`, which reqwest does not cap. Every other transport stops at `transport::MAX_BODY` (64 MiB). The unlocker returns the *target page* inside a JSON string (`parse_response` → `j["body"].as_str().to_vec()` = a second copy), and `cache_put` base64-encodes it to disk (a third), so the page's author picks the size ×3. `read_capped` drains `chunk()` by `chunk()` and refuses past the same cap with `BypassFail::Network("… exceeded the … cap")`.

## 2. `Proxy::parse` errors echo the password (LOW)

`proxy.rs:205/:213/:217`: `"proxy: bad auth in {s}"` etc., where `s` is `user:pass@host:port`. That lands in a `FetchError::Http` which can surface in a tool result or a log line; the `Debug` impl (`:148-162`) was deliberately redacted for exactly this reason and the parse errors were the missed sibling. The messages now name the expected shape; the bad-port variant quotes only the port token.

## Tests

- `an_oversized_unlocker_response_is_refused_at_the_body_cap` — a local server answering 65 MiB through a real `reqwest::Response`; RED with the cap disabled: `Ok(65 MiB)`.
- In `transport::proxy` tests: three malformed entries (`bad port`, `no port`, `no colon in auth`) whose error text must not contain `s3cret`.

`cargo fmt --check`, `clippy --all-targets` clean, `nextest` 1333/1333 (soak excluded locally). CHANGELOG has an Unreleased entry.


dondai44423:
Folded a review fixup: the bad-port message quoted the port token, but in a shape without `@` (or with the order swapped, `host:port@user:pass`) that token is the password. The message names the shape now, and the test covers both extra shapes.

PR #275, opened 2026-09-20
fix(pdf): the fetch tool parses PDFs off the runtime worker, and a document is read to 500 pages

mnaza:
Two halves of one starvation vector on attacker input, from a sweep of `pdf/*` and the MCP fetch path at 4.2.9 (the rest of the PDF engine was clean: size gate before pdfium, 80 M-px bitmap guard, page-object clamp, bounded meta/bookmark buffers, every pdfium handle behind a `Drop` guard, no `unwrap` on pdfium results, OCR already capped and on the blocking pool).

## 1. `web_fetch` parses a PDF inline on the tokio worker (MEDIUM)

`fetch_tool.rs` calls `extract::extract` inline at every site — `:1161` (tier-1 result), `:1218` (links re-extract of the same body), `:1455` (pagination parts), `:1671` (unlocker body). For a PDF that is `pdf::parse` → `engine::load_document`: `FPDF_LoadPage`, a 96 dpi rasterization, `FPDFText_LoadPage` and the layout/table/fusion passes for **every page**, synchronously, holding pdfium's global lock. Consequences: that worker does nothing else for the parse; the call's own `deadline_ms` wrapper cannot fire because the future never yields inside it; other tool calls scheduled on that worker wait. `crawl/mod.rs:1153-1170` already did this right (`spawn_blocking` + 300 s timeout) — the fetch tool was the missed sibling.

Fix: `extract::extract_off_worker(body, ct, url, opts)` — a PDF body goes to the blocking pool under `PDF_EXTRACT_BUDGET` (300 s), HTML goes straight through (fast, already bounded by the walk caps). The four fetch sites and the crawl use it; the crawl's inline copy is gone, and a typed `ExtractError::Failed` replaces the `BadSelector("extract timed out …")` shape it used for the timeout.

## 2. No page cap (MEDIUM-LOW → makes half 1 hours instead of seconds)

`engine.rs:767` iterated `0..FPDF_GetPageCount`; only OCR pages were bounded. `pdf_max_mb` (100 MB) admits ~100 k near-empty pages, and an *empty* page measured 100–300 ms here (the rasterization plus passes), so a document whose every page is blank was hours of work for a note saying "scanned". `MAX_PAGES = 500` (carried in `LoadOpts.max_pages` so a test can set it lower); `RawDoc.pages_read` records what was read, `page_count` stays the true count, and `parse` adds a note: `document has N pages; the first 500 were read (per-document page cap)`.

## Tests

- `pdf::tests::pages_past_the_cap_are_reported_not_read` — a generated well-formed 8-page PDF (correct xref) through `parse_with(.., 5)`: `page_count == 8`, note names "8 pages" and "first 5 were read"; a 3-page document under the cap has no note; `MAX_PAGES` is pinned at 500. RED with the cap bypassed: the note is absent.
- `extract::tests::extract_off_worker_parses_pdf_off_the_runtime_and_html_inline`.

`cargo fmt --check`, `clippy --all-targets` clean, `nextest` 1334/1334 (soak excluded locally). CHANGELOG has an Unreleased entry for both.


dondai44423:
Folded a review fixup: the ghost retry and the wayback snapshot still called `extract` inline, and both can carry PDF bytes (`is_binary` admits them), so they go through `extract_off_worker` too.

PR #276, opened 2026-09-20
fix(extract): a pre-parse nesting gate, own-row table scans, a JS-data scan budget, reddit depth caps

mnaza:
Four extraction DoS shapes from a sweep of `extract/*` at 4.2.9, three of them any-host, all on the tokio worker that runs `extract`. Each verified with a RED test on FETCH_HEAD.

## 1. html5ever is O(n²) in nesting depth, and nothing capped the depth (MEDIUM, any host)

The tree builder scans the stack of open elements on every tag (`in_scope`, implied end tags — `html5ever-0.39.0/src/tree_builder/mod.rs:1084-1165`), as the spec writes it. Browsers cap parser nesting at 512; html5ever has no cap. Measured on the main `extract` path: 1 000 nested `<div>` → 0.3 s, 2 000 → 1.05 s, 4 000 → 3.4 s. A 1 MiB page of them (200 000 deep) is **hours**, and every host can serve one.

Fix: `extract::nesting::max_nesting` — a linear scan that counts only what the parser's stack actually keeps: void elements, self-closing tags, raw-text bodies (`script`/`style`/…), comments, quoted attributes and sibling-closed elements (`p`, `li`, `td`, …) are excluded, so unclosed runs of those on real pages do not count. Past `MAX_NESTING = 4096` the body is refused as "not a document" (`ExtractError::Failed`) — on the decoded HTML path before any adapter, and in the feed path (unknown elements stay on the stack too). 50 000 nested `<div>` is refused in < 1 s; a wide page of the same byte count extracts as usual.

## 2. Nested tables: quadratic, and the rows merged (MEDIUM, any host)

`blocks.rs:569/605` `is_prose_table`/`table_block` used descendant selects (`el.select("tr")`, `tr.select("td,th")`). At every nesting level the outer row "held" every nested cell, each scanned to the bottom by `text_size`, then `walk` recursed a level and did it again. 1 000 nested tables (18 KB) took minutes (RED: `TIMEOUT 60 s`); and a nested table's rows became rows of the outer table. `own_rows`/`own_cells` read the table's own rows (directly or via `thead`/`tbody`/`tfoot`) and the row's own cells: 0.3 s, and the nested table stays inside its cell (`| a |` rows no longer leak into the outer table).

## 3. `jsdata::find_blobs` rescans per marker (MEDIUM-LOW, any host)

A known-global marker (`window.__NEXT_DATA__ = `) with no value behind it makes `extract_js_value` scan to the end of the document, then `from` advances one char and the next marker repeats it: N × document. 512 KiB of markers was minutes (RED: `TIMEOUT 60 s`). The scan has a budget of 4 × the document length; a real page has a handful of markers each scanned once, and a value at the end of a marker run is still found.

## 4. Old-reddit renderer: no depth cap, descendant list selects (LOW, host-gated)

`render_comment`/`render_md`/`render_list` recursed without a cap (`blocks.rs` and `math.rs` have `MAX_DEPTH`) and `render_list` used descendant selects (quadratic, every nested item rendered once per ancestor). 3 000 levels — under the gate, so it does reach the adapter — overflowed a 1 MiB stack (RED: `SIGABRT … has overflowed its stack`). `MAX_DEPTH = 100` threaded through all three; the list renderer reads own items and own nested lists.

## Note on #275

This branch adds the same `ExtractError::Failed` variant as #275 (same lines, same text), so whichever merges second applies cleanly.

`cargo fmt --check`, `clippy --all-targets` clean, `nextest` 1339/1339 (soak excluded locally). CHANGELOG has an Unreleased entry for all four.


dondai44423:
Folded fixups. A self-closing slash on a non-void tag was left out of the gate, but the HTML parser ignores it: `<div/>` nested to a real DOM depth of 32 000 while the scan reported 0, and extraction stayed quadratic there. It counts now, and the gate test carries that shape. The 3.4 s figure at 4 000 didn't reproduce on this box (parse ~110 ms; the quadratic shows from 8 000: 8 000 0.4 s, 16 000 1.6 s, 32 000 6.6 s), so the committed comments and the changelog use the re-measured numbers. The gate test moved above the table section so #275's eof append merges cleanly.

